PR #115, opened 2015-01-14
Exit the game prematurely if DIABDAT.MPQ or Diablo.exe cannot be loaded

alexey-lysiuk:
Also prevents crash on exit in this case


wheybags:
Apart from that one issue, looks good to me. Once that is fixed I'm happy to merge.
Also, hello, haven't seen you around before I don't think.
Feel free to stick around, help is always appreciated!

PR #124, opened 2015-05-12
Launcher

konopka90:
Finally I made first version of launcher! 
Additional requirements: 
   - Library Qt 5.x. 

Features to do in the future:
   - Add nice images
   - Implement rest of checkboxes/comboboxes
   - Add more messageboxes with errors/warnings etc.

Unfortunately, launcher was tested only on Ubuntu 15.04 with Qt 5.4 so please check on other platforms if it's working properly. Incomplete CMake file and handling paths to files/dirs can be a problem. 

Whats' more I added new component: Settings. It is very simple class with hardcored paths, I think there is no need (so far) to improve its functionality. We should use this class not only in Launcher but also in Freeablo.


konopka90:
yaa cmake problem :3

wheybags:
Awesome, this is really great!
I'll pull this into a branch and fix the travis cmake issues then merge it.
Also gonna have to get it working on windows, as it adds a new dependency (uuugh)

konopka90:
Qt is really heavy, I know this pain :hamster:

wheybags:
merged!

PR #127, opened 2015-05-13
Monsters can spawn on level entry stairs fix

konopka90:


konopka90:
hmm stairs are 3x3 grid ?

wheybags:
This should do, the other tiles in the grid aren't passable, so it won't spawn there anyway

PR #128, opened 2015-05-15
Minor fix for audio

konopka90:
On my env without enabled sounds freeablo crashes because Audio::playSound always returns -1.


wheybags:
Huh, wouldn't have expected that

PR #129, opened 2015-05-18
Video Player

konopka90:
Hi ! 
I've got next feature for you - SMK Video Player.
Additional dependency:
   - FFmpeg library

For now player is very poor - it can only render videos in original dimensions. Loading times should be optimised. What's more sometimes there are weird segmentation faults when application deallocates memory - I don't know really why. Maybe the reason is that I'm working on VM.. 
Anyway I'm going to work on it.
Tested only on Ubuntu.

Greetings,
Patrick.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ Find_Package(PythonLibs 2 REQUIRED)
 Find_package(LibRocket REQUIRED)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5Core REQUIRED)
+find_package(FFMPEG REQUIRED)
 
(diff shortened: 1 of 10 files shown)


wheybags:
holy shit, nice!
Didn't expect this :p
Will have a closer look at this tomorrow

konopka90:
Yesterday I was very tired so code is messy ;p 
Things to do:
   - eliminate seg faults and memory leaks
   - reduce similar allocations - move buffers to static scope (?) , explicite deallocation at the end of application life
   - speed-up loading videos - still don't know how
   - load and play audio from .smk
   - not sure about solution with boost::timer

PR #130, opened 2015-05-19
Video Player v0.2

konopka90:
Additional dependency:
   - FFmpeg library

Features:
   - Eliminated mem leaks and seg faults 
   - Enhanced code clarity
   - Custom video width and height!

To do:
   - Audio
   - Read FPS from file (on the other hand all videos in MPQ has 15FPS)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ Find_Package(PythonLibs 2 REQUIRED)
 Find_package(LibRocket REQUIRED)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5Core REQUIRED)
+find_package(FFMPEG REQUIRED)
 
 if (Qt5_POSITION_INDEPENDENT_CODE)
     SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
@@ -39,7 +40,7 @@ if(NOT WIN32)
     endif ()
 endif()
 
-find_package(Boost COMPONENTS chrono date_time filesystem system random thread program_options atomic python REQUIRED)
+find_package(Boost COMPONENTS chrono date_time filesystem system random thread program_options atomic python timer REQUIRED)
 
 if ( NOT SDL2_FOUND )
     message ( FATAL_ERROR "SDL2 not found!" )
@@ -60,6 +61,7 @@ include_directories (
     ${PYTHON_INCLUDE_DIR}
     ${ROCKET_INCLUDE_DIR}
     ${Qt5Widgets_INCLUDE_DIRS}
+    ${FFMPEG_INCLUDE_DIRS}
     "${PROJECT_SOURCE_DIR}/components/"
     "${PROJECT_SOURCE_DIR}/extern/StormLib/src/"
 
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -85,6 +87,8 @@ link_libraries (
     ${PYTHON_LIBRARIES}
     ${Boost_LIBRARIES}
     ${Qt5Widgets_LIBRARIES}
+    ${FFMPEG_LIBRARIES}
+    swresample
```
wheybags: this is unnecessary, and bypasses cmake

components/CMakeLists.txt
```diff
@@ -85,3 +85,10 @@ add_library(Settings
 )
 target_link_libraries(Settings ${Boost_Libraries})
 set_target_properties(Settings PROPERTIES COMPILE_FLAGS "${FA_COMPILER_FLAGS}")
+
+add_library(Video
+    video/video.h
+    video/video.cpp
+)
+target_link_libraries(Video ${FFMPEG_LIBRARIES} swresample)
```
wheybags: same as before

components/video/video.cpp
```diff
@@ -0,0 +1,252 @@
...
+          }
+
+        // Free the packet that was allocated by av_read_frame
+        av_free_packet(&packet);
+    }
+
+    av_free(bufferRGB);
+
+    avformat_close_input(&formatContext);
+    avformat_free_context(formatContext);
```
wheybags: avformat_close_input frees this already, and sets the pointer to NULL

components/video/video.cpp
```diff
@@ -0,0 +1,252 @@
...
+      mNumFrames(0)
+{
+}
+
+Video::~Video()
+{
+    int size = mSprites.size();
+    for(int i = 0 ; i < size; i++)
+    {
+        SDL_DestroyTexture(reinterpret_cast<SDL_Texture*>(mSprites[i]));
```
wheybags: This should be handled in the render component, nothing but that should know sdl even exists

wheybags:
Had to add 
```
#ifndef AV_CODEC_ID_SMACKVIDEO
    #define AV_CODEC_ID_SMACKVIDEO CODEC_ID_SMACKVIDEO
#endif

#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(55,28,1)
    #define av_frame_alloc  avcodec_alloc_frame
#endif
```
to get it to build on my system, but apart from that, and the few issues I mentioned above it looks great!

Would you like me to merge now, or do you want to take care of the to do items first?
Also, sorry it took so long to get to this, haven't had the time to look at it properly until now.

konopka90:
I'm trying to do better implementation - this solution consumes a lot of memory so please don't merge now. Loading times in FullHD are confusing (all frames are loaded at once). I was trying to do little hack: load frames in small resolution and then scale it in renderer but scaling in renderer doesn't look good - it should happen in "sws_scale" function with proper filter (e.g. bilinear or bicubic). I think about typical producer-consumer problem: producer produces frames in the fly and consumer displays these frames.

wheybags:
Yeah, I was thinking you only need one SDL_texture, which you can just
update every frame, but this would do for now

CMakeLists.txt
konopka90: I remember that I had problem with linking on Ubuntu without these lines but I'm not sure for now, I'll check it.

PR #131, opened 2015-05-26
Video Player v0.3

konopka90:
Hi,
I rewrote video player including your suggestions. Still I have problem with audio, in general I got solution but sometimes it hangs up in one loop and I can't figure out how to solve this problem. 
So far we can merge player without sound.

Changes:
   - Fast video loading
   - Proper video scaling with filter

Greetings,
Patrick.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,8 +87,6 @@ link_libraries (
     ${PYTHON_LIBRARIES}
     ${Boost_LIBRARIES}
     ${Qt5Widgets_LIBRARIES}
-    ${FFMPEG_LIBRARIES}
-    swresample
     Cel
     FAIO
     Levels
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ Find_Package(PythonLibs 2 REQUIRED)
 Find_package(LibRocket REQUIRED)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5Core REQUIRED)
+find_package(FFMPEG REQUIRED)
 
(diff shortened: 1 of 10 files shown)


wheybags:
One more small ffmpeg incompatability, and removing swresample was required, diff below:
```
diff --git a/components/CMakeLists.txt b/components/CMakeLists.txt
index 6147736..d2ef39e 100644
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -90,5 +90,5 @@ add_library(Video
     video/video.h
     video/video.cpp
 )
-target_link_libraries(Video ${FFMPEG_LIBRARIES} swresample)
+target_link_libraries(Video ${FFMPEG_LIBRARIES})
 set_target_properties(Video PROPERTIES COMPILE_FLAGS "${FA_COMPILER_FLAGS}")
diff --git a/components/video/video.cpp b/components/video/video.cpp
index 058dadd..91cdb99 100644
--- a/components/video/video.cpp
+++ b/components/video/video.cpp
@@ -14,6 +14,10 @@ namespace Video
     #define av_frame_alloc  avcodec_alloc_frame
 #endif
 
+#ifndef AVPixelFormat
+    #define AVPixelFormat PixelFormat
+#endif
+
 AVCodec * Video::sCodec;
 AVFrame * Video::sFrame;
 AVFrame * Video::sFrameRGB;
```

Is that swresample line actually required on your system?
If so, could you try to load it in using a proper cmake find script, not just adding it in manually as you did?

Apart from that, getting occasional `Error reading from file` messages in console, probably just a small mistake in file reading.

konopka90:
I suggest to copy-paste and customize implementation from OpenMW, it would be better solution.

PR #137, opened 2015-07-13
DiabloExe::BaseItem, DiabloExe::Prefix, FAWorld::Inventory, draggable widgets and more

mloccy:


apps/freeablo/faworld/inventory.cpp
```diff
@@ -0,0 +1,243 @@
...
+            case eqRING:
+                if(item.item.equipLoc == eqRING)
+                {
+                    if(!ringSlot) //ringSlot=0 => left hand ring
+                    {
+                        if(&this->leftRing == &emptyItem) return true;
+
+
+                    }
+                    else if(ringSlot) //ringSlot=0 => right hand ring
```
wheybags: should be ringSlot=1 in the comment?

resources/python/draggable.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from docmanage import DocManager
+import inspect
+import pprint
+class DraggableWidget(object):
+    def __init__(self, document, xoffset, yoffset):
```
wheybags: document isn't actually being used here

mloccy:
This is now ready to be pulled I think

mloccy:
There's still a bug in libRocket that means that textures wont show in the inventory, as far as I can tell librocket is not picking up background-image selectors when elements are dynamically created in python. Not sure if this would happen with non-cel images.

PR #147, opened 2015-08-09
Fix warnings for gcc and partially for VS

konopka90:


konopka90:
This should be done during refactoring main.cpp and other files.

PR #151, opened 2015-08-13
Settings refactor

konopka90:
Seems to work.


components/settings/settings.cpp
```diff
@@ -28,52 +35,155 @@ bool Settings::loadSettings()
...
+        return false;
     }
 
-    return child.get();
+    return true;
 }
 
-template<class T>
-void Settings::set(const char* name, T value)
+bool Settings::loadFromFAFile(const std::string & path)
```
wheybags: This function is never used

konopka90: Done

konopka90:
Ready to merge.

PR #156, opened 2015-08-16
Replace boost with std library

konopka90:
Removed boost libs: atomic, random, chrono, date_time and thread.
Please check if everything works fine.


mloccy:
Builds and runs perfectly for me

PR #159, opened 2015-08-18
Item refactor

mloccy:


mloccy:
Well I need to pass the chosen class to the python script so it can select the correct class. And I'm trying to remove references to the character names that were used in Diablo.

PR #163, opened 2015-08-23
GUI for console

konopka90:
Hi, maybe it is too early for this feature but I made simple GUI for console. For now it launches everywhere except main menu.
   - Fixed height 600px
   - Show/hide under ~
   - Accept commands and adds to visible area
   - Arrow up - show previous command in input
   - Arrow down - show next command in input

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,7 @@ link_libraries (
     StormLib
     Audio
     Settings
+    Script
 )
 
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
--- a/apps/freeablo/CMakeLists.txt
+++ b/apps/freeablo/CMakeLists.txt
@@ -39,6 +39,8 @@ add_executable(freeablo
     fagui/console.cpp
     fagui/commandhistory.h
     fagui/commandhistory.cpp
+    fagui/autocomplete.h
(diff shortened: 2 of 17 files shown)


wheybags:
Looks good so far, but ofc it doesn't actually do anything yet.
Could you make it so it just functions as a python shell?
If you don't want to yet, that's fine, we can add that in later.

wheybags:
Oh, also, could you make it pass key events through to the base rml when its hidden? like https://github.com/wheybags/freeablo/blob/master/resources/gui/bottommenu.rml#L76

(currently if you open + close console, then press esc, it will not pause, as the key events still go to the console)

konopka90:
Ok i will experiment this week

konopka90:
Key events and python shell works properly. Added shortcuts for quit/pause/unpause/load game/save game in console (just type f.e. "quit" and press enter").

wheybags:
Cool, looks great, just one last nitpicky thing I noticed is printing newlines, eg: print "asasd\nadasd", it will just print a space where the newline should be.

konopka90:
Yup, also there's problem with multiline output. Resize window to ~400px x 800px and type command "freeablo.quit(1,2)" :3

konopka90:
Please check if everything is good. In the future I'm gonna to add auto-complete feature.

konopka90:
Autocomplete feature delivered.

PR #164, opened 2015-08-24
Set window to fullscreen if user requests it.

rohit-n:


wheybags:
Hey, would you like to make it render on the selected screen too?

rohit-n:
I'll look into that.

PR #171, opened 2015-09-06
Load Diablo.exe from user-specified path.

rohit-n:


wheybags:
Sorry it took so long to get to this!

PR #172, opened 2015-09-06
Player factory

konopka90:


wheybags:
sorry it took so long to get to this.
This refactor was needed, actor class needs some more refactoring even I'd say.

PR #176, opened 2015-09-25
New font Freeablo in all menus

konopka90:


wheybags:
Nice!
Just one issue I can see for now, the W in the screenshot below, you can see the next glyph peeking in at the right
screenshot from 2015-09-26 16-56-44

konopka90:
Fixed. One more issue - this menu looks different on my pc. Pentagrams are inline with text, did you change anything in your build ? 

diff

wheybags:
Nope, nothing. Even just tried deleting my settings-user.ini there

konopka90:
Not so good, it sounds like there are differences in librocket on Windows and Linux build.

wheybags:
My screenshot was from Linux, was yours?

konopka90:
Yes, mine was from Linux too

sheepo99:
To be fair, I'd prefer to have a new font created from scratch altogether, so we could finally have expanded Latin/Russian/Greek characters. This would allow us to finally translate the game to other languages, which would be an interesting addition to Freeablo.

wheybags:
merged it, I suspect the issue is just related in general to ui scaling, and it works fine on another computer that I'm using fro development atm

PR #184, opened 2015-12-16
Attack bug

konopka90:
On my build I couldn't attack from "behind" of monster and left side of monster but I was able to attack from right and bottom side. I found odd behaviour in computing distance to enemy, this patch should fix it. 

Unable to attack:
unable

Able to attack:
able


konopka90:
One more thing - do you have segmentation fault when you attack from left side ? I don't know why but when I attack from left side game crashes but when I attack from top/bottom/right and rest of sides it works fine o.O
Please check before/after patch.

Crash:
crash

PR #185, opened 2015-12-16
GUI for Single player creator

konopka90:
Character creator added:
+ select hero menu
+ select class menu
+ enter name menu
+ invalid name screen

\+Main menu display fix.


wheybags:
when i select create a new character, it will always tell me i entered an invalid name.
Also, the invalid name menu has an "ok" option, that i can only select by clicking, hitting enter doesn't work

konopka90:
Yes, there is still no validation, i dont know how diablo validates name but it's not important for now. What is odd it says that you cannot use space but.. you can :)

wheybags:
Was that last commit supposed to fix the issues I described?
Just checked and it's still there

konopka90:
> Was that last commit supposed to fix the issues I described? 

No, i didn't read it before commit. Please check now.

wheybags:
lgtm

PR #190, opened 2015-12-24
sync npcs

wheybags:


apps/freeablo/faworld/playerfactory.cpp
```diff
@@ -6,59 +6,56 @@
...
-    mStats->recalculateDerivedStats();
+    player->mInventory.collectEffects();
+    stats->recalculateDerivedStats();
 
-    return mPlayer;
+    return player;
 }
 
-void PlayerFactory::createWarrior()
+ActorStats* PlayerFactory::createWarrior(Player* player, const DiabloExe::CharacterStats& charStats) const
```
konopka90: Would be more natural if createWarrior/createRogue/createSorcerer returns object of class Player rather than ActorStats. Maybe try to remove parameter "player", create ActorStats inside this func and return Player.

wheybags: That makes sense. Honestly, I think it makes more sense to do all this stuff inside the Player constructor, but the pattern in the codebase is that *Stats are created outside of the Actor and then passed in, which I think should be changed at some point to being created in the constructor

PR #192, opened 2015-12-27
Hello my friend, stay a while and listen!

konopka90:
Basic mechanism for dialogues 
+ Opening dialogue
+ Playing sound

Go to the story teller or smith and click on him!


wheybags:
When the dialogue menu opens, it opens at the top left of the screen. Maybe the middle of the screen would be better?
It's fine if it blocks access to all other ui, iirc it does in diablo as well

konopka90:
Yes, I will try to do it as in D1 - show dialogue menu on right side of screen. And if dialogue menu opens then GUI with stats/spells etc. will hide.

konopka90:
Someone can review and I will go further

konopka90:
Dialogues block input to gui now like in D1. You can check it now.

wheybags:
Cool, looks good!

PR #194, opened 2015-12-27
Compilation fix

konopka90:


wheybags:
That'll teach me to merge without running it first :p
(it was a one line diff though ;_;)

Issue #197, opened 2016-01-01
mpqtool: 'std::logic_error'


gbudny:
Hi

I noticed this binary file in the Freeablo directory, but I can not run it:

./mpqtool
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct null not valid
Aborted (core dumped)

uname -a
Linux use-Macmini 3.13.0-63-generic #103-Ubuntu SMP Fri Aug 14 21:42:59 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux

konopka90:
Hi, it expects one argument - path to file. Try ./mpqtool PATH_TO_DIABDAT.MPQ

wheybags:
Nope, actually it expects two args, the first is a path inside the
mpq(source), the second is a local path (destination).
It will extract the source to the destination. It depends on having
DIABDAT.MPQ in the working dir

gbudny:
I am not sure if this command is correct:

ls
apps         CMakeCache.txt       components   exedump   launcher  resources
celview      CMakeFiles           DIABDAT.MPQ  extern    Makefile
celview.ini  cmake_install.cmake  Diablo.exe   freeablo  mpqtool

 ./mpqtool DIABDAT.MPQ .
Segmentation fault (core dumped)

wheybags:
no, it doesn't take the mpq as an argument, jsut two file paths as I said
in the last post.
eg: ./mpqtool "levels/l2data/l2.min" l2.min
and then you'll have l2.min extracted from the mpq in your current dir

gbudny:
Thank you.

It works now.

./mpqtool "levels/l2data/l2.min" l2.min
levels\l2data\l2.min

 ls
apps         CMakeCache.txt       components   exedump   l2.min    mpqtool
celview      CMakeFiles           DIABDAT.MPQ  extern    launcher  resources
celview.ini  cmake_install.cmake  Diablo.exe   freeablo  Makefile

I have to admit this isn't a user-friendly application.

Do you plan to create a GUI for this program?

wheybags:
No, it's not really meant for users, just a convenient little program for
developers if they need to extract a file or two and aren't on windows (on
windows you can use ladik's mpq editor)

gbudny:
Thanks for the answer.

Ragora:
Sounds like it should at least enforce the two-argument requirement and dump a small help in the event argc != 3.

PR #201, opened 2016-01-01
Add parameter message & short explanation to MPQTool

Ragora:
In reference to #197. It seems a little improper to not have an error message explaining what the program does and what input it takes when the input requirements are not met. Also it looks like my tabbing may be off (looked right in my editor, damn it !)


konopka90:
Looks good, just change tabs to spaces.

Ragora:
Should be good now.

PR #210, opened 2016-01-17
Fix fullscreen setting not being read correctly.

rohit-n:


konopka90:
@gbudny I think that your problem is related to other issue, something with menu rendering. Does this problem exist also in build from master branch with/without fullscreen? 

Anyway, PR works good, can be merged.

konopka90:
@gbudny In file https://github.com/wheybags/freeablo/blob/master/apps/freeablo/engine/enginemain.cpp in line 72 add   
``` 
Render::updateGuiBuffer(NULL);
``` 
and please tell me if it helps.

gbudny:
@gbudny

This issue exists in builds from all branches in the with/without fullscreen mode.

I added this line, but it did not help me to solve this issue.

wheybags:
Works fine here (on osx), so I'll merge, sorry for the delay.
@gbudny that issue seems unrelated to this patch, but still worrying.
Is it still happening to you?

gbudny:
I checked this new version of Freeablo, and I have the same issue on Mac OS X.

Issue #104, opened 2014-12-02
level generation for dungeon level 4


wheybags:
see #93 for example

PR #220, opened 2016-04-11
Level 4 generation

theotime:
Hello,
This patch should give the level 4 generation basics (#104).

Few notes:
* There are no doors in L4, so I replaced the door by floor before the tiling of the level and let the tiling function do its job to add *WallEnd. It gives quite good results.

* Tell me if I'm wrong, but with the L4, there is a glitch with the current generation algorithm, Corners are slightly different for L4. See the difference between L2 & L4 in the screenshots above.

joinl2
joinl4

For L4 there is two different corners, one for simple corner and one for a tri-join corner, _insideRightCorner_ should have the tile ID 16 but also the tile ID 13 version to handle this precise case.
So, the algo has to be modified to fix that. It will dive in it.

Tell me if it's okay for you.

Thanks!

### DIFF
--- a/apps/freeablo/falevelgen/levelgen.cpp
+++ b/apps/freeablo/falevelgen/levelgen.cpp
@@ -581,7 +581,7 @@ namespace FALevelGen
     // Helper function for adding doors
     // Iterates over all blocks on a wall, and adds doors where necessary, looking at what is in the direction
     // indicated by add (1 or -1) to determine if a door is needed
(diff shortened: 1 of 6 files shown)


wheybags:
Nice!
Yeah, the tilesets are all super awkward :s (and if you look closely the other levels aren't generated perfectly anyway :p)
At some point I imagine someone will do a full pass over level generation and fix all the annoying little edge cases but for now this is great!
Also glad to see a contribution from someone new, welcome!

theotime:
Thank you!

PR #222, opened 2016-04-16
Bottommenu pausemenu bug

Exairnous:
fixes a couple problems with the pause menu and makes it more like diablo's


wheybags:
What exactly is this patch supposed to do? It seems I can click on the bottom menu buttons while the game is paused and it will unpause? Is this what it was meant to do?

Exairnous:
Yes.  This patch does a couple things, for one it makes the pause menu act in mostly the same way as the original Diablo (the behaviour you noted);  in Diablo the char screens stay up and the menu actually covers them, but given that this is already a major change I didn't want to go to far.  The pause menu also had a nasty bug where if you clicked outside the menu it would lose focus and under the right conditions if you pressed ESC it would end up that the game unpaused but the menu didn't disappear; this is now fixed.

wheybags:
I'm all for the bug fix, but not a fan of having the bottom menu usable through the pause menu unfortunately :s.
Any chance you could open a new PR with just the bugfix?

Exairnous:
That's the way it is in the original Diablo, but I'll be glad to isolate just the bugfix and re-upload it.  Should be done in a couple of days or less; if it isn't just poke me again :)

Issue #204, opened 2016-01-07
Drag menu bug


konopka90:
Do you have the same effect?
https://gfycat.com/MisguidedMellowAvocet

gbudny:
I can confirm that I have the same issue in Freeablo.

konopka90:
This issue can be closed

PR #223, opened 2016-04-16
Drag dialog fix

Exairnous:
Fixes Drag menu bug #204, repositions character info screens over their buttons, and makes them drag only from their top borders.


gbudny:
Thanks.

It works, but I noticed some issues:

[Warning] - Syntax error parsing property declaration 'position: inherit;' in resources/gui/inventory.rml: 9.
[Warning] - Syntax error parsing property declaration 'position: inherit;' in resources/gui/character.rml: 11.
[Warning] - Syntax error parsing property declaration 'position: inherit;' in resources/gui/spells.rml: 9.
[Warning] - Syntax error parsing property declaration 'position: inherit;' in resources/gui/quest.rml: 9.

Traceback (most recent call last):
  File "<string>", line 2, in Event_9aee8d10
NameError: global name 'docmanage' is not defined
tick time exceeded by 52ms

https://gist.github.com/gbudny/5bf9b6a2c8200100c5da1d3dcb998c55

wheybags:
Looks mostly good, except for the issues gbudny raised.
Also it seems like it's not always passing mouse events through to base.rml properly.
In the long run I'd like to avoid the current mechansim for distinguising ui clicks from game clicks.
What I'd like to do is use "colour picking", where we create an offscreen render target, clear it to white, then draw the ui with a special shader that just outputs black for any pixel that has alpha > 0.
Then we can get the click coords and check whether the pixel at that point is black or white.

Buuuuut that's all a ways off for now, would require porting the renderer to opengl (which will happen at some point, hopefully for 0.5?), so for now we should make sure the current mechanism is working.

Exairnous:
@gbudny I think I've fixed your issues, if not let me know.
@wheybags the colour picking method for mouse clicks sounds neat! (first time I've heard of it :) ).  With regards to the base.rml not getting mouse clicks, the only place I found trouble was right at the top of the bottommenu (because the bulbs stick out) and I'm pretty sure that's not caused by my patch.  Could you tell me exactly where it was failing?

gbudny:
**@Exairnous**

Thanks

It works. 

I didn't see any warnings or errors.

wheybags:
Ok, I can't remember what the focus issue was, nor reproduce it, and it should all just go away once I do color picking anyway, so this looks good :P

Exairnous:
Sounds good :)

PR #234, opened 2016-06-04
Add support for version 1.09b of Diablo.exe.

mewmew:
Note, the file size of the diablo.exe binary of v1.09 and v1.09b
is identical, and more importantly they both have the exact same
addresses for every function and global variable; thus a symlink 
to `v109.ini` is sufficient.

ref: https://freeablo.org/forum/viewtopic.php?f=4&t=9#p116


wheybags:
Been meaning to do this for a while, thanks!

PR #245, opened 2016-09-27
Add inventory hotkey

Fuco1:
Hit `i` to open the inventory dialogue.  Hit again to hide it.

toggleDialogue:
In case we want to control the dialogues with a hotkey, it is convenient
to open when closed and close when opened.  This simplifies the client
code checks for the state of the dialogue.

### DIFF
--- a/apps/freeablo/engine/engineinputmanager.cpp
+++ b/apps/freeablo/engine/engineinputmanager.cpp
@@ -159,6 +159,9 @@ namespace Engine
             case QUIT:
                 actionAsString = "Quit";
                 break;
+            case INVENTORY:
+                actionAsString = "Inventory";
+                break;
             case NOCLIP:
                 actionAsString = "Noclip";
                 break;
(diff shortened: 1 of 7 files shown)


wheybags:
It seems you're creating a second document for the inventory - press i to open inv and then move it over, then click the inventory button and you'll get a second inv screen. Also, it's very slow to open your inv - there's a fre hundred ms pause when you press i. I think this is because you are creating and destroying the document instead of hiding and showing it.

Fuco1:
Right, `Close` unloads it and `Hide` just hides.  Got it.

One thing I don't get is, what exactly is `mDocument` supposed to be in the gui manager?  Because clearly you can have multiple documents shown (character and inventory), but there is only one variable.  Is it the "last" worked on document?  I don't suppose there is one guimanager per document.

wheybags:
I believe that is the root document that sits underneath all others. It's
used to detect whether a click was on the game or the GUI.

wheybags:
*bump* :3

Fuco1:
Let's close this since it is irrelevant after the renderer rewrite.

Issue #247, opened 2016-11-05
RAII IO


wheybags:
The current io system is implemented in components/faio, as a series of functions similar to the fopen/tread c io functions.
It would be nice to have a clean RAII class that would use these functions to open the file in its constructor, and close it in it's destructor.
I'd still want it's member functions to be similar to dread etc, as opposed to going the route of c++ streams though

This could be a good task for someone new to the codebase ehi wants to get their feet wet.

SGBon:
I might give it a look. Do you want to keep the same interface with the friend functions? And if so should they be adapted as wrappers that call internal member functions or still contain all the code?

wheybags:
I wouldn't remove the existing functions - I'd just implement the wrapper on top of them, so the wrapper would still have an FAFile* member variable.

SGBon:
Alright then

pengyz:
I will do something about it too.

PR #248, opened 2016-11-07
try fix issule#247(RAII IO)

pengyz:
fix for #247:
1. add **FAFileObject** class for RAII wrapper of the original C style  FAIO functions.
2. Change all callers to use the new wrapper besides the **FAIOFileInterface**(it can just use the original c style io  functions).


pengyz:
On, I will fix the compile warnings since the vs don't give me it.

wheybags:
Basically looks good! Just one issue is that you used literal tabs in a few places, whereas the rest if the project uses 4 spaces for indentation. If that's fixed then I think I'm happy to merge.

pengyz:
ok，it's just a auto format by the visual studio. I'll fix it.

wheybags:
Great, thanks!
Sorry for the delay

PR #155, opened 2015-08-15
Path finding based on A* algorithm

konopka90:
I created implementation of A\* algorithm and sample program with few testcases. Interface "Level" is similar to original class "Level" so it will be easy to adopt this solution. In my opinion it works quite good and fast. For now I have to add cost checking - diagonal vs non-diagonal move. Regarding algorithm if user selects unreachable location it will find closest point and path for this location.

Remarks: works only with C++11.
Reference: http://www.redblobgames.com/pathfinding/a-star/implementation.html#cplusplus

Issue #146, opened 2015-08-08
NPC ai


wheybags:
basic pathfinding towards the player + attacking

konopka90:
#155

wheybags:
closed by #263

PR #249, opened 2016-11-16
Findpath integrate

pengyz:
Try To fix #155 (integrate find path with  A* algorithm writtened by @konopka90).
It is also about the #146 (NPC ai)

I have only tested on the player(beacuse it's easy by click the mouse.). Also I think it will be fine on the npcs(maybe we need a ai script for the monster and npc.).
I intergate findpath module to the class actor by add a method like this:
            void findPath(GameLevelImpl* level, std::pair<int32_t, int32_t>& destination);
Also,the calss Position is changed to restore the path points that FindPath::find returned.

I think I need a code review to improve it. :-)


wheybags:
Cool, great to see someone having a go at this!
I can't seem to use it though, when I click behind the house at the town spawn the player just walks at the wall like normal.
Is it just not done yet? How should I test it?

pengyz:
Really? Oh, it's a bad news. 
I'll test it again and try to fix it . I hate bugs. :-(

pengyz:
I make a bug fix now. I'm not sure if your problem about "the player just walks at the wall like normal" is fixed. There has a bug before that the role will walk a step on the impassable tile but it will move back automaticlly immediately. I have fixed the problem I found. I'm glad if you can test it again. :-)

wheybags:
Great, it works for me now!
Code looks ok, I'm not sure I understand all of it just yet, but nothing looks outrageously horrible :p
If you're happy to go ahead, I can merge this now, unless you'd like to do some more work on it?

One thing I did notice is that it doesn't respect npcs when pathing, but that's not a huge issue.
Also, going to put a video of this up on the blog, as it's pretty cool :3

wheybags:
Oh, and again, sorry for the delay in reviewing this :(

pengyz:
I known this problem...
This findpath code is only according to the map passable,
```C++
bool GameLevel::isPassable(int x, int y) const
{
    return mLevel[x][y].passable();
}
```

I don't know how to check the Actor's collision size...
Alos as you say,It's not a huge issue.I'm looking at the Entity Component System now. You have told me that you prefer to use it in freeablo. I'm working for it now,but the progress is really slow. :-(
More time needed for doing this.

I think you can merge it if it seems ok. I'll fix the issue sometimes later,Just after I have understood the Actor  and know how to check the actor collision size.. :-)

wheybags:
@pengyz Cool, I've got some free time now over christmas to work on freeablo, so if you've lost interest in it then I'm happy to start working on that. Another thing we really need is proper stats calculation, just in case you were looking for something :p

Anyway, my plan is to work on porting the renderer to actual opengl for now, followed by either ECS or stats, or maybe switching gui library.

PR #250, opened 2017-02-25
OpenGL rendering

wheybags:
-librocket doesn't work, but I want to get rid of it anyway ~~(probably for myGui + lua)~~ Nuklear! https://github.com/vurtun/nuklear
-probably won't be merged until this is done


CMakeLists.txt
```diff
@@ -49,6 +49,15 @@ endif ()
 
 set (CMAKE_CXX_STANDARD 11)
 
+# Use C++11
+if (CMAKE_VERSION VERSION_LESS "3.1")
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+        set (CMAKE_CXX_FLAGS "--std=gnu++11 ${CMAKE_CXX_FLAGS}")
```
Fuco1: Any reason not to use c++14? I've had it set locally ever since I've started working on this project and everything seems to work just fine.

wheybags: Travis is still using gcc 4.6.3, I'm not sure how much of c++14 is supported in that?

Fuco1: Oh, I see.  No big deal I guess, as far as I can tell we're not using any new features.  But it's good to preemptively use newer standards if possible.

Issue #149, opened 2015-08-10
refactor rendering


wheybags:

mloccy:
Might as well fix the renderer so that it only renders what's in the viewport if you're going to refactor it.

konopka90:
Big issue is about rendering bitmap fonts in GUI - as we know development of librocket stopped long time ago. This PR https://github.com/libRocket/libRocket/pull/34 is impossible to merge. It needs to be reimplemented from scratch what will introduce a lot of changes in librocket. I have no idea how to solve this problem - try to improve librocket, maybe librocket + CEGUI or totally remove librocket from the project or something else.

ChefKeeper:
Hello,
I'm looking forward to contribute !
What exactly needs to be done or redone with rendering ?
(I'm not saying I'll do it, I just need to understand before deciding)

wheybags:
Hi!
Sorry for the delay, meant to reply sooner.
It's great to hear that you want to contribute. However, refactoring rendering is probably not the best place to start (it's components/render/sdl2backend.cpp that's the problem, the nastiest magic-number fueled code in the project :p).
If you're looking for something ambitious to do though, why not try to integrate pathfinding to make the monsters actually go towards the player?
@konopka90  actually wrote an implementation of the A* pathfinding algorithm, but it hasn't been integrated with the freeablo character/level system and is sitting in a PR (#155).

To give some pointers on how to get started: all entities (players, npcs and monsters) in the world are instances of FAWorld::Actor. The actor class has an update function that is called every frame. In this update function, there is code that will move the npc towards a location held in the mDestination member. The position is held in the mPosition member which is an instance of the FAWorld::Position class. Position defines a location on the game grid, accounting for the fact that an actor can be in between two adjacent tiles. To do this it essentially stores two grid locations, the current and next tile coord (but the next coord is just stored as a direction, since it is always adjacent to the current), and then also where the actor is between them as an int between 0 and 100.

The current code will just move in a line toward the actor's mDestination, but it would be nice if we could have it pathfind towards it instead. It would also be nice if hostile NPCs (FAWorld::Monster) would set their mDestination to players locations when inside a certain radius of them.

So that ended up being a bit of an essay, but anyway, hope you're interested in that job but if not it's a nice description for someone to pick up anyway :p

ChefKeeper:
Hey no problem we all got jobs ;)
All right I'll look to integrate his PR sounds fun :) 
[...]

PR #251, opened 2017-02-26
Rendering Refactoring

Predelnik:
So I started up with the very simple goal in mind to fix viewport centering at 640x480 which was annoying me but since I had to figure out how most of the rendering code works anyway and since I saw issue #149 I decided to rewrite most of it in hopefully more clear way.

1. So the first commit should mostly keep rendering algorithm functionally exactly the same, except maybe be a bit more efficient at certain steps. But it still draws objects in two loops ordered like it was originally.

2. The second commit I'm not actually sure of. I do think that rendering out-of-map tiles should be fixed but I haven't figure out how to find the appropriate tile texture for that. Taking the first one seems to work fine but obviously looks a bit crude. But since it works maybe it's okay to leave it until future better implementation.

3. In the third commit I've changed drawing order to horizontal lines (with new code it wasn't that hard) which may do somehow affect performance but I don't think it's super relevant. It fixed actually very rare and minor glitch for me which I may make gif of if you'd like. The basic problem lies in moving of Actors (even for now it's only player actor which is moving) - since we take into account only starting point of actor's movement in rendering it could lead to things being drawn incorrectly while it still between source and destination tiles.

So for future reference - drawing n moving actors absolutely correctly may require sorting visible actors by their current y screen coordinate and drawing them between corresponding lines of objects in that order. But I'm not sure that it's needed for good visualization or even implemented correctly in Diablo itself so it requires some testing in the future.

PS. I also added some helping structs like `Point`, `Vector`, `Tile` to help me with specifying operations. It's fine to replace them with more generally used versions in the future if there's any, I just didn't find anything similar. Struct `Tile` should technically be in World namespace but in my future commits I had to distinguish between halves of a tile for targeting and it's not required for most of the objects, so maybe `World` needs another `Tile` struct. Anyway I do not think that using `std::pair` for that is a correct decision.

### DIFF
--- a/components/render/render.h
+++ b/components/render/render.h
@@ -36,28 +36,6 @@ namespace Level
 
 namespace Render
 {
-    // Difference between two points
-    struct Vector
-    {
-      int32_t x;
-      int32_t y;
-
-      Vector operator* (const double &scalar) const
-      {
-        return {static_cast<int32_t> (x * scalar), static_cast<int32_t> (y * scalar)};
-      }
-    };
-
-    // Point in pixels
-    struct Point
-    {
-      int32_t x;
-      int32_t y;
-      Point operator+ (const Vector &v) const { return {x + v.x, y + v.y}; }
-      Point operator- (const Vector &v) const { return {x - v.x, y - v.y}; }
-      Vector operator- (const Point &other) const { return {x - other.x, y - other.y}; }
-    };
-
     // Tile mesasured in indexes on tile grid
     struct Tile
     {
--- a/components/render/render.h
+++ b/components/render/render.h
@@ -155,7 +155,7 @@ namespace Render
     SpriteGroup* loadTilesetSprite(const std::string& celPath, const std::string& minPath, bool top);
     void drawLevel(const Level::Level& level, size_t minTopsHandle, size_t minBottomsHandle, SpriteCacheBase* cache, LevelObjects& objs, int32_t x1, int32_t y1, int32_t x2, int32_t y2, size_t dist);
 
-    Tile getClickedTile(const Level::Level& level, size_t x, size_t y, int32_t x1, int32_t y1, int32_t x2, int32_t y2, size_t dist);
+    Tile getClickedTile(size_t x, size_t y, int32_t x1, int32_t y1, int32_t x2, int32_t y2, size_t dist);
 
     void clear(int r = 0, int g = 0, int b = 255);
 }
--- a/components/render/render.h
+++ b/components/render/render.h
@@ -36,6 +36,34 @@ namespace Level
 
 namespace Render
 {
+    // Difference between two points
+    struct Vector
+    {
+      int32_t x;
+      int32_t y;
+
+      Vector operator* (const double &scalar) const
+      {
+        return {static_cast<int32_t> (x * scalar), static_cast<int32_t> (y * scalar)};
+      }
+    };
+
+    // Point in pixels
+    struct Point
+    {
+      int32_t x;
+      int32_t y;
+      Point operator+ (const Vector &v) const { return {x + v.x, y + v.y}; }
+      Point operator- (const Vector &v) const { return {x - v.x, y - v.y}; }
+      Vector operator- (const Point &other) const { return {x - other.x, y - other.y}; }
+    };
+
+    // Tile mesasured in indexes on tile grid
+    struct Tile
+    {
+      int32_t x;
+      int32_t y;
+    };
     /**
      * @brief Render settings for initialization.
      */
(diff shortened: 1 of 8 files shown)


wheybags:
Great, it sounds good based on your description, but I haven't actually read the code yet, I'll try to do a proper review soon.
One slightly awkward coincidence is that I have a branch open that replaces SDL rendering with pure opengl, so I'll need to merge this code into that branch, which probably won't be much fun but hey, it's a good problem to have :p

Also, I'm so sorry for the state of sdl2backend.cpp :p
It was truly atrocious

wheybags:
Oh, also looks like Travis found a few trivial problems, so if you could fix those that would be nice

Predelnik:
Yeah I've seen your commit about changing SDL rendering hopefully it shouldn't even intersect very much because I kept basic `drawSprite` function exactly the same.

Hopefully I fixed Travis errors. It's nice to be checked by it since I can't compile on Linux )

components/render/render.h
```diff
@@ -36,6 +36,34 @@ namespace Level
...
+    {
+      int32_t x;
+      int32_t y;
+      Point operator+ (const Vector &v) const { return {x + v.x, y + v.y}; }
+      Point operator- (const Vector &v) const { return {x - v.x, y - v.y}; }
+      Vector operator- (const Point &other) const { return {x - other.x, y - other.y}; }
+    };
+
+    // Tile mesasured in indexes on tile grid
+    struct Tile
```
wheybags: I'd say it would be fine to just use Vector for this. Also, could combine Point and Vector.
And it should probably be in the Misc component, not here. But I agree, using std::pair is nasty

components/render/sdl2backend.cpp
```diff
@@ -818,203 +820,117 @@ namespace Render
...
+    // as simple as possible
+    static Point tileTopPoint (const Tile &tile) {
+      return {(tileWidth / 2) * (tile.x - tile.y), (tile.y + tile.x) * (tileHeight / 2)};
+    }
 
-        int32_t x = xPx1 + ((((float)(xPx2-xPx1))/100.0)*(float)dist);
-        int32_t y = yPx1 + ((((float)(yPx2-yPx1))/100.0)*(float)dist);
+    // this function simply does the reverse of the above function, could be found by solving linear equation system
+    // it obviously uses the fact that ttileWidth = tileHeight * 2
+    Tile getTileFromScreenCoords(const Point& screenPos, const Vector& toScreen)
```
wheybags: Could you explain what's going on here? Was my implementation of this function ludicrously over-complicated or is this doing something different now?

components/render/sdl2backend.cpp
```diff
@@ -818,203 +820,117 @@ namespace Render
...
+        spriteSize(sprite, w, h);
+        auto point = pointBetween (start, finish, dist);
+        auto res = point + toScreen;
+        drawAtTile(sprite, res, w, h);
+    }
 
-        levelX = xPx1 + ((((float)(xPx2-xPx1))/100.0)*(float)dist);
-        levelY = yPx1 + ((((float)(yPx2-yPx1))/100.0)*(float)dist);
+  constexpr auto bottomMenuSize = 144; // TODO: pass it as a variable
+  Vector worldToScreenVector(const Tile& start, const Tile& finish, size_t dist)
```
wheybags: nitpicking, but dodgy indentation here

components/render/sdl2backend.cpp
```diff
@@ -818,203 +820,117 @@ namespace Render
...
-        getMapScreenCoords(level, x1, y1, x2, y2, dist, levelX, levelY);
+    void drawLevel(const Level::Level& level, size_t minTopsHandle, size_t minBottomsHandle, SpriteCacheBase* cache, LevelObjects& objs, int32_t x1, int32_t y1, int32_t x2, int32_t y2, size_t dist) {
+      auto toScreen = worldToScreenVector({x1, y1}, {x2, y2}, dist);
+      SpriteGroup* minBottoms = cache->get(minBottomsHandle);
+      auto isInvalidTile = [&](const Tile &tile){ return tile.x < 0 || tile.y < 0 || tile.x >= static_cast<int32_t> (level.width()) || tile.y >= static_cast<int32_t> (level.height());};
 
-        //TODO clean up the magic numbers here, and elsewhere in this file
+      // drawing on the ground objects
+      drawObjectsByTiles (toScreen, [&](const Tile &tile, const Point &topLeft){
+          // Fill invalid tiles with ground, it looks ok but it's probably better to have something else than zero-eth sprite here
```
wheybags: Eventually I'd like to rig up infinite outward spread of some tileset-appropriate value, but for now this is fine.

components/render/sdl2backend.cpp
Predelnik: I do believe it technically does the same thing but likely +/- some constant (x,y) which is irrelevant because we center everything on a player in the end anyway. I don't know if it was 'overcomplicated' but it contained at least some ifs and this version is pure arithmetics with division by 64 which is obviously optimized to >> 6. So it should perform nicer.

components/render/render.h
Predelnik: Generally I think more types are better because of type-safety (e.g. not allowing to pass Tile where argument is Point for example). I'm fine with combining Point and Vector and moving them to Util, I do think that Render::Tile should stay though because later I added possibility to disntiguish left and right halves of a Tile which are distinguished in original game for targetting by mouse.

wheybags: OK, that makes sense so

wheybags:
Sorry it took so long to get back to you, I've been swamped at work.
Looks good to me, and it works, so I'll merge it now.
I'd still like an explanation of how getTileFromScreenCoords works though, as atm I just don't understand it (although I can see that it works, so I'm ok to merge it :p)

PR #252, opened 2017-02-26
Fix some type casting warnings.

Predelnik:
Mostly this commit straightforwardly adds casts in the places where Visual Studio complains about `loss of information`. It shouldn't do any functional change for the most part. However I changed `Dun` class to store `std::size_t` instead of `int16_t`  because it had just a ton of casts backwards and forwards from `size_t`.  And I don't think it's necessary to store numbers in the same way they were stored in binary if we have to cast them each time anyway.


wheybags:
lgtm

PR #253, opened 2017-02-26
Make dungeons generated on demand rather than on start.

Predelnik:
Description is pretty obvious. I just want to note that it's tremendously speeds up start of program in debug and I do think it's important.

### DIFF
--- a/apps/freeablo/faworld/world.cpp
+++ b/apps/freeablo/faworld/world.cpp
@@ -90,7 +90,7 @@ namespace FAWorld
 
         for (int32_t i = 1; i < 17; i++)
         {
-            mLevels[i] = FALevelGen::generate(100, 100, i, mDiabloExe, i - 1, i + 1);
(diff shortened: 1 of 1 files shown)


apps/freeablo/faworld/world.cpp
```diff
@@ -151,7 +151,14 @@ namespace FAWorld
 
     GameLevel* World::getLevel(size_t level)
     {
-        return mLevels[level];
+       auto p = mLevels.find (level);
+       if (p == mLevels.end ())
+         return nullptr;
+       if (p->second == nullptr)
+         {
```
wheybags: dodgy indentation here, otherwise lgtm

Predelnik:
I've fixed that. Btw is that ok if I do `git push -f` like here or is it better to add requested changes through new commits?

wheybags:
That's fine, my policy is generally force-pushing prs is fine, but not
master

wheybags:
bump :v

PR #254, opened 2017-02-27
Add approximate .clang-format style to help with more consistent formatting

Predelnik:
Sadly it's not perfect due to the fact that clang-format missing some features like omitting space around binary operators. But I think it works fine as a reference point. I don't recommend using it globally though but applying it locally may help people like me who tend to indent incorrectly often.


wheybags:
Cool, I'd like to maybe apply it globally once, and get the nasty messy diff over and done with, then have travis enforce it.
I'm not going to do that right now though, so I'll just merge this for now

PR #256, opened 2017-03-11
Actor hover

Predelnik:
Some details:
1. Status bar is mostly obvious, just showing actor's name for now. Sadly it required to throw in some hackery i.e. changing rml element attributes in runtime in C++ due to the curious nature of how Diablo aligns text. I came out with html solution but librocket didn't manage to render it correctly sadly.
2. Hover highlighting algorithmically is very easy, we just build texture filled with required color for each 4-connected pixel except exactly black (because black is shadow). However implementation is not obvious, in these commits I build mirror white texture `mask` for each texture to be drawn under the actor with requested color. However later for items I actually made generating highlighted texture with specific color by request (using `&higlight=#XXXXXX`) (Due to the fact that it was needed in librocket code). So there's a bit of logic/code duplication. In general I think the second way is better but it requires some refactoring to implement it for actors. I think later on mask texture may be removed, but for now it works so it's probably ok.
3. How targetting works is found purely by experiment. It seem to follow the same pattern for actors even if actor seems small. Targetting pattern is red hexagon for actor staying on tile a:

However for some triggerable items possibly future inspection is required.
4. Some fixes for targetting/walking which actually of little matter since I've written mostly new code for all this when writing item pick-up (future commits). The only thing to note is that dead bodies should probably in the future stop being actors since they can a share tile with living actor.

### DIFF
--- a/apps/findpath/main.cpp
+++ b/apps/findpath/main.cpp
@@ -46,6 +46,11 @@ namespace FAWorld
             return gameMap[y][x] == 0 ? true : false;
         }
 
+        bool isPassableFor(int x, int y, const Actor *) const override
+        {
+            return isPassable (x, y);
+        }
+
     private:
         int gameMap[MAP_SIZE][MAP_SIZE];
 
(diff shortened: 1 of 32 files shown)


Fuco1:
> The only thing to note is that dead bodies should probably in the future stop being actors since they can a share tile with living actor.

I have addressed this in #262.  I think we should keep the actor instances "alive" even when dead in case features such as resurrection would be implemented (for monsters or players alike).

Predelnik:
This now is fully included in #293

PR #259, opened 2017-04-26
Fix markdown header

danyshaanan:


afritz1:
There are some other places in the Readme with the same problem. You should add those to your pull request, too. I was going to do it myself, but I saw that you already started it!

danyshaanan:
@afritz1 Right! I've went through all the markdown file and fixed whatever I could find. (mainly headers and code snippets formatting).

wheybags:
Nice, thanks!
Did GitHub change how they format markdown? All those headers used to work fine with no space after the #. Same thing happened to the readme of our private repo at work.

danyshaanan:
Yeah, I've seen this before. Don't know if they changed how they format it. [Only git will tell](https://github.com/github/markup).

PR #262, opened 2017-06-24
Add support for rendering of multiple actors at the same tile

Fuco1:
Before the change only one actor per tile was supported in the renderer.  This prohibited some features such as walking over dead enemies (they would flicker out of existence because the player would be drawn instead).

I have changed the data structure from holding a single entity to a vector of entities.  I still haven't figured out how to properly layer them but it seems that in vast majority of cases the player is layered above the dead creature.

The vector is then rendered in order so when we make sure dead enemies are in front and the player comes last it will be properly layered and displayed.

I have also made a change to allow walking over dead enemies.

Imgur

Imgur

### DIFF
--- a/apps/freeablo/farender/renderer.cpp
+++ b/apps/freeablo/farender/renderer.cpp
@@ -223,7 +223,7 @@ namespace FARender
                 {
                     for(size_t y = 0; y < mLevelObjects.height(); y++)
                     {
-                        mLevelObjects[x][y].valid = false;
+                        mLevelObjects[x][y].clear();
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


apps/freeablo/farender/renderer.cpp
```diff
@@ -233,13 +235,16 @@ namespace FARender
...
+                    Render::LevelObject levelObject = {
+                        std::get<0>(state->mObjects[i])->isValid(),
+                        std::get<0>(state->mObjects[i])->getCacheIndex(),
+                        std::get<1>(state->mObjects[i]),
+                        position.next().first,
+                        position.next().second,
+                        position.mDist
+                    };
+
+                    mLevelObjects[x][y].push_back(levelObject);
```
wheybags: Could maybe have some variable on actors that indicates the frame they arrived at their current position, and then sort the list by that variable?

Fuco1: That sounds like a great idea, I'll see what I can do.

It would be good to also somehow rewrite `actorMapRefresh` and friends to not wipe & reinsert every time, then we could just cache them once in some map and only insert and delete actors who move.  The dead ones don't need refreshing especially since they are not animating either.

wheybags: Yeah, it would but for now it seems to work ok. I figured I'd come back to it if it was ever a performance problem

PR #263, opened 2017-06-24
Add basic AI

Fuco1:
Monsters will now engage if you are closer than 5 tiles (hardcoded for now).  They will path and walk towards the player, then attack.

The AI scripts are separate classes and can be attached to actors.  This way one AI behaviour can be reused for multiple monsters or monsters can switch the AI scripts based on various conditions (for example on low HP monster could switch to a "coward" script and run away).


apps/freeablo/faworld/monster.h
```diff
@@ -10,18 +11,30 @@ namespace DiabloExe
...
+            void attachAI(AI* ai) {
+              mAi = ai;
+            };
+
+            bool attack(Actor * enemy);
+            void update(bool noclip, size_t ticksPassed);
+            bool canIAttack(Actor * actor);
 
         private:
+            AI * mAi;
```
wheybags: I would implement this on Actor, then maybe having a special "playerAI" that allows an actor to be controlled by keyboard and mouse IE, the player character. In this case, it might be better to rename AI to behaviour, but that might be overkill. I'd definitely move the code to Actor though, anyway

wheybags:
Also, general notw on all these PRs thank you so much! It's great to see someone come along and start submitting stuff. If you're working on more stuff, or looking for some stuff to work on, do let me know

apps/freeablo/faworld/monster.h
Fuco1: I've added the `NullAI` for this reason mostly, it simply does nothing so actor can initialize to that and monsters can then initialize some other behaviour.

I'll move it up, it makes sense since other entities will also have their AIs (like towns people or cows :D though I don't remember if they wandered around).

Fuco1:
I've submitted a PR something like 9 months ago (it is still open there with some issues :D) then somehow lost track of free time.  Now things are finally getting back to order so I'm quite pleased to get my hands on this.  Thanks for the encouragement!

Hop on the gitter sometime, I'll try to hang around and ask questions :)

wheybags:
I should get notifications on gitter, but my phone was off for most of this weekend :p
Also, will probably just have to close your old pr unfortunately, I've since started redoing the GUI not using librocket, so it will break when that's merged

Fuco1:
I moved the AI up to the actor class and renamed it to behaviour.  When/if we implement some message bus/event system we could set up behaviour for the player as you suggested to consume the UI events and direct the actor to do things.  That would be nice and decoupled and dandy :)

Fuco1:
The windows build is failing with 

```
c:\projects\freeablo-windows-build\freeablo\apps\freeablo\faworld\behaviour.cpp(34): error C2220: warning treated as error - no 'object' file generated [c:\projects\freeablo-windows-build\build\apps\freeablo\freeablo_lib.vcxproj]
```

I'm not quite sure what's the matter :/

apps/freeablo/faworld/actor.h
```diff
@@ -206,6 +213,7 @@ namespace FAWorld
 
         protected:
             GameLevel* mLevel = NULL;
+            Behaviour* mBehaviour;
```
wheybags: should be = nullptr here, was getting a segfault without

Fuco1: Oh silly C++ does not initialize to null pointer by default? :O

Fuco1:
We could probably only run the AI on monsters that are "about to be visible" for the player, updating the entire map seems pointless indeed.  ISTR that diablo 2 works that way (maybe even original diablo); when you start it in ridiculous resolutions and you can see way beyond what would normally be accessible the monsters just stay there doing nothing.

The engine in general seems rather unoptimized which I guess is actually a good approach: make things work first, make things work fast later.

I've tried using callgrind on the vision branch because I got massive slowdown as well but wasn't really able to find anything useful (20% of the time was spent in stormlib but that's probably the bias of having too few game loops and so the majority of the time was spent on init).  I've solved it by recomputing the vision only when the tile the actor's on changes, so it does it once a sec (roughly) instead of 120 times :)

wheybags:
Merged current master into this branch. Responsiveness is a bit better since it's not loading sprites all the time, but it's still a bit slow.
Also, there seems to be a bug where sometimes enemies don't play their death animation, and just "die" standing up :v

I'm off for the next month or so, so I've got time to work on this bug+perf issues if you want, but it's your PR so I'll leave it to you if you want (plenty of other stuff for me to work on :p)

I'd like to finally release v0.4, and I'd love to have this in there

apps/freeablo/faworld/monster.cpp
```diff
@@ -19,11 +20,46 @@ namespace FAWorld
...
+        if(enemy->isDead() && enemy->mStats != nullptr)
+            return false;
+        isAttacking = true;
+        Engine::ThreadManager::get()->playSound(FALevelGen::chooseOne({"sfx/misc/swing2.wav", "sfx/misc/swing.wav"}));
+        setAnimation(AnimState::attack, true);
+        mAnimPlaying = true;
+        return true;
+    }
+
+    bool Monster::canIAttack(Actor * actor)
```
Fuco1: This method probably be removed now with the "factions" code merged in.

Fuco1:
I think the bug you are talking about is simply caused by lags, then the ticks jump forward by the whole animation length.  I've noticed that too.

Feel free to add features, open source is about collaboration after all :)

What do you think about the optimization proposed above (about only processing part of the enemies?).

We should also think about the networking and how this would work out... because I use random and so every player can get different monster updates.  Is there some framework for "shared" random (via seeds or so).  Or how would we synchronize the monster positions/actions/ai... I've looked a bit at the serialization code but don't get it yet.

I think it would be best to run the same simulation on every client (which is what happens now I think) so we should probably distribute the AI scripts and "updatability" of the monsters.

wheybags:
Ok, I'll have a go at fixing that then. As for only processing some npc, yeah, I think that will be essential really. I also think we can do something like every odd frame process npcs with id%2 == 0, every even id%2 == 1 to further cut down on updates. And, also, can probably make pathfinding much faster too, I don't think there's been any attempt to optimise it yet

Aso for networking, it's not a big deal if client and server get a bit out of sync, the way it works now is that the same simulation is run (but not *strictly* the same, floating point differences and other stuff could happen), and actor state is fixed up periodically by just overwriting local state with server state.
But yeah, it would be nice if monsters didn't decide to walk different directions too often, not sure how much work it would be to set up a shared randomness, it would mean we'd have to start taking care of a lot of stuff we currently don't (like order of updates for actors).

wheybags:
Started refactoring animation to fix that bug, and things got a bit out of hand :vBeginning to regret starting it on this branch, but oh well, it's done now :s
Will need to refactor movement too, it's a bit wacky, then I can try making it not slow as hell

apps/freeablo/faworld/actor.h
```diff
@@ -62,26 +91,25 @@ namespace FAWorld
                   const std::string& dieAnimPath=""
                   );
 
-            virtual void update(bool noclip, size_t ticksPassed);
+            virtual void update(bool noclip);
```
Fuco1: What's the reasoning behind removing the ticks?  I think it's useful to have that in update methods to have some sort of time flow.

apps/freeablo/faworld/behaviour.cpp
```diff
@@ -34,8 +34,10 @@ namespace FAWorld
         return nearest;
     }
 
-    void BasicMonsterBehaviour::update(size_t ticksPassed)
+    void BasicMonsterBehaviour::update()
     {
+        Tick ticksPassed = World::get()->getCurrentTick();
```
Fuco1: This might get pretty annoying when we will try to write tests.  Especially these behaviour classes look like good candidates for testing because they are pretty stateless (or at least they handle the state internally).

Fuco1:
It's odd though, on my computer (which is a shitty 2core celeron 1.2ghz) I don't really notice it being extremly slow.  I would say it goes about the same speed on this branch as it used to on the master at the time.

I would like to look at the pathfinding code to optimize that a bit.  Some years ago I was quite obsessed with optimizing A* :)  I'll do that as a separate patch though.

Fuco1:
> Will need to refactor movement too, it's a bit wacky, then I can try making it not slow as hell

What you mean by this exactly?  Do you have something concrete in mind?  Maybe write some proposal somewhere, maybe I could look at it.

apps/freeablo/faworld/actor.h
wheybags: Well, I needed to use world->getCurrentTick elewhere anyway, so I figured it was better to be consistent

wheybags:
There's pathfinding data in the Pos class, which seems a bit off to me. Also, it was just crazy complicated. I've simplified it a bit already, but could do a good deal more. Nothing too concrete, just a better separation between position, movement and pathfinding

apps/freeablo/faworld/behaviour.cpp
```diff
@@ -0,0 +1,68 @@
...
+                // we want to move some more
+                if (ticksPassed - mLastActionTick > ((size_t)rand() % 300) + 100) {
+                    // if 1 let's move, else wbehaviourt
+                    if (rand() % 2) {
+                        auto next = mActor->mPos.current();
+                        do {
+                            next = mActor->mPos.current();
+                            next.first += ((rand() % 3) - 1) * (rand() % 3 + 1);
+                            next.second += ((rand() % 3) - 1) * (rand() % 3 + 1);
+                        } while (!mActor->canWalkTo(next.first, next.second));
```
Fuco1: This should also probably have some counter with an upper bound in case no tile around is walkable (which might happen if the monster gets teleported or something)

apps/freeablo/faworld/behaviour.cpp
```diff
@@ -38,14 +38,22 @@ namespace FAWorld
...
+
+            int32_t dist = FAWorld::squaredDistance(nearest->mPos, mActor->mPos);
+
+            if (dist <= 25) // we are close enough to engage the player
+            {
                 mActor->mDestination = mActor->mPos.mGoal = nearest->mPos.current();
             }
-            // if no player is in sight, let's wander around a bit
-            else if (!mActor->mPos.mMoving) {
+            else if (dist >= 200) // just freeze if we're miles away from anyone
```
Fuco1: This is a radius of about 14 tiles, which might not be enough on large enough resolutions.

For the future maybe we could extract a method into the world class or somewhere that would check if a tile is "visible" by any player and then we could use that.  This is more like just a todo note than anything, it's fine for now.

wheybags: Yeah, I was thinking of doing something like that myself.
Should at least up that value, though

wheybags:
Now all I need to do is get muliplayer working again after I butchered it for this pr :v

PR #264, opened 2017-06-25
Use pushdown state-machine to handle actor state

Fuco1:
The code is still hairy and I used some hacks like friend classes (for now) to enable the first round of refactoring.

Next steps will be to move all the behaviour to their own states and define proper interface to talk to the actor.

So far there is only the "BaseState" (which will turn abstract eventually) doing all the work old `Actor::update` was doing except handling the attack state, I have added a state for that as an example of how simple it can be :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,7 @@ link_libraries (
     StormLib
     Audio
     Settings
+    StateMachine
     Script
     Serial
 )
--- a/apps/freeablo/CMakeLists.txt
+++ b/apps/freeablo/CMakeLists.txt
@@ -17,6 +17,11 @@ add_library(freeablo_lib # split into a library so I can link to it from tests
     farender/spritemanager.cpp
     farender/spritemanager.h
 
+    faworld/actor/basestate.cpp
+    faworld/actor/basestate.h
+    faworld/actor/attackstate.cpp
+    faworld/actor/attackstate.h
+
     faworld/actor.cpp
     faworld/actor.h
     faworld/monster.h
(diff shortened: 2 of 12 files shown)


wheybags:
I like the idea in principle, can you explain how it works/will work in practice? Also, it seem you've got some link errors on windows, according to appveyor

Fuco1:
It's based somewhat loosely on combination of ideas from this article: http://gameprogrammingpatterns.com/state.html It's a nice read so you might want to read it anyway.  It explains it very well.

Basically it is a [pushdown automaton](https://en.wikipedia.org/wiki/Pushdown_automaton).  Every state is a class with an update method which can return optionally a new state or nothing (then it remains in that state).  Additionally to the returned state you specify one of the three operations: pop, push, replace.  

With these you control the "stack" of state scripts.  This is super helpful as it allows us to remember where we came from (i.e. on top of the standard FSM we have a notion of history through the stack).  So I can simply change states as usual with replace or I can push a new state, then do some stuff in it and then the state can pop itself off and the actor returns to whatever they were doing previously.

An example can be a creature that wanders around or does some other business, but when the player approaches it switches to a "flee" state.  When they are far enough the state pops itself and the creature reverts to whatever they were doing before.

wheybags:
Link error on windows is because there's no .cpp files in StateMcahine component. If you add a dummy empty cpp file in there, then it should link fine

wheybags:
Ready for merge? I'm happy to merge now if you are

Fuco1:
@wheybags  Yep.  I'll integrate the other branches with this when merged (there have been some conflicts in the `Actor::update` method) and then we can continue refactoring the updates into separate states.

wheybags:
Cool, and thanks again!

PR #265, opened 2017-06-25
Add factions to support better ally/foe detection.

Fuco1:
The member variable `mEnemy` on the actor is insufficient because actors can be both players and monsters and possibly other factions (angles, townsfolk...).  I have added factions and resolution of who is an enemy based on that, so that monsters see the player and townsfolk as enemies and they see monsters as enemies.

### DIFF
--- a/apps/freeablo/faworld/actor.h
+++ b/apps/freeablo/faworld/actor.h
@@ -14,7 +14,7 @@
 #include "position.h"
 #include "gamelevel.h"
 #include "world.h"
-
+#include "faction.h"
 
 namespace Engine
 {
@@ -39,7 +39,7 @@ namespace FAWorld
             hit,
             ENUM_END // always leave this as the last entry, and don't set explicit values for any of the entries
         };
-    }    
+    }
     class Actor : public NetObject
     {
         STATIC_HANDLE_NET_OBJECT_IN_CLASS()
@@ -64,7 +64,7 @@ namespace FAWorld
             virtual FARender::FASpriteGroup* getCurrentAnim();
             void setAnimation(AnimState::AnimState state, bool reset=false);
             void setWalkAnimation(const std::string path);
-            void setIdleAnimation(const std::string path);            
+            void setIdleAnimation(const std::string path);
             AnimState::AnimState getAnimState();
             bool findPath(GameLevelImpl* level, std::pair<int32_t, int32_t> destination);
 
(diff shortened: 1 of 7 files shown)


apps/freeablo/faworld/faction.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include "actor.h"
+
+namespace FAWorld
+{
+
+    bool Faction::canAttack(Faction* other)
+    {
+        // Attacker -> Attackee
+        static bool ruleTable[FactionType::ENUM_END][FactionType::ENUM_END] =
```
wheybags: How about we just have every faction attack other factions for now? Later on we could add in a system that allows dynamically adding alliances, if we need to.

apps/freeablo/faworld/actor.h
```diff
@@ -209,7 +210,7 @@ namespace FAWorld
 
             bool mIsDead = false;
             bool mCanTalk = false;
-            bool mIsEnemy;
+            Faction* mFaction;
```
wheybags: I would prefer if this variable wasn't a pointer

apps/freeablo/faworld/faction.h
```diff
@@ -0,0 +1,48 @@
...
+    class Faction
+    {
+    public:
+        Faction(FactionType faction): mFaction(faction) {};
+
+        bool canAttack(Faction* other);
+
+        static Faction* hell()
+        {
+            return new Faction(FactionType::hell);
```
wheybags: similar to above comment, I'd just use a stack variable here

apps/freeablo/faworld/actor.h
Fuco1: I had it as a regular variable before but changed it to a pointer as I had some troubles with initialization.  I agree it would be better, I will try to change the code back. (I don't remember what the problem was :D I've been hacking on this since 10 AM till now hehe)

apps/freeablo/faworld/faction.cpp
Fuco1: Sounds good, but then we need to have townsfolk in some faction and we don't want the player attack them.  So we can file them under hero, or just have two alliances now "heaven" and "hell" for people and monsters respectively.

wheybags: Yeah, two factions is fine for now I'd say

Fuco1:
@wheybags I've fixed the issues above.

wheybags:
Thanks!

PR #285, opened 2017-08-20
Implemented dummy (for now) panels for inventory, spells etc.

Predelnik:
The first commit blocks events handled by nuklear so that character walk etc. is not triggered when we click on various panels.

The second - adds dummy panels. The only problem with panels is that in original game they are drawn under the bottom panel (it's visible by health/mana bubble) and nuklear seems to force active windows on top so their order is changing if they are clicked. Probable workaround of attaching `NK_WINDOW_BACKGROUND` to them somehow breaks event handling so in the future it needs to be investigated further.


wheybags:
Nice, thanks!

PR #286, opened 2017-08-21
Bugfix for my previous PR about panels

Predelnik:
Apparently hiding windows is not enough to treat them as non-active so they have to be closed. I filed an issue about this to nuklear but for now I think this should work.

### DIFF
--- a/apps/freeablo/fagui/guimanager.cpp
+++ b/apps/freeablo/fagui/guimanager.cpp
@@ -124,9 +124,10 @@ namespace FAGui
             return 0;
             }(), screenH - 125 - invTex->getHeight(),
             invTex->getWidth(), invTex->getHeight());
-        nk_flags flags = NK_WINDOW_NO_SCROLLBAR;
-        nk_fa_begin_image_window(ctx, panelName (panelType), dims, flags, invTex->getNkImage(), op);
-        if (*panel (placement) != panelType)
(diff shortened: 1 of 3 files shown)


Predelnik:
I also added one-line nuklear fix for making background windows work as I think they should. (I asked about it in nk's issue tracker but got no answer for now). 

Plus I added nuklear define which makes button trigger on mouse release as it works in Diablo with its 6 main button. Though if you consider other things buttons too, sometimes they trigger on mouse down, that's why it's quite unfortunate that it's determined by define but well there's always workarounds.

wheybags:
I would very much like to avoid going down the route of customising nuklear for this project.
That said, this pr works and you haven't got a response on the nuklear issue tracker yet, so I will merge but I'd like to see this issue resolved upstream.
To clarify, the issue that this solves is that creating a window with NK_WINDOW_BACKGROUND will block input from going to the currently active window at the point when the background window is created?
I'm surprised there's been no response yet, @vurtun is normally very responsive, I'm sure he'll reply in the next couple of days.
Also just adding in a link to the nuklear issue, in ase anyone is reading this and curious: https://github.com/vurtun/nuklear/issues/503

vurtun:
Hey,
it took some time but I finally fixed the issue. Sadly @Predelnik s  fix of just removing the flag does not work under all circumstances but I hacked in a solution to make it work.

PR #288, opened 2017-08-26
Make inventory display and drag&drop work.

Predelnik:
Please merge #286 before.


resources/gui/bottommenu.rml
```diff
@@ -66,21 +66,22 @@ docmanage.manager.loadDoc(docmanage.manager.InventoryFile)
 docmanage.manager.loadDoc(docmanage.manager.CharacterFile)
 docmanage.manager.loadDoc(docmanage.manager.SpellsFile)
 docmanage.manager.loadDoc(docmanage.manager.QuestFile)
-        </script>
```
wheybags: why did you modify this?

extern/nuklear/nuklear.h
```diff
@@ -19916,6 +19917,22 @@ nk_widget_is_mouse_clicked(struct nk_context *ctx, enum nk_buttons btn)
 }
 
 NK_API int
+nk_widget_is_mouse_click_down(struct nk_context *ctx, enum nk_buttons btn, int down)
```
wheybags: Can you open a PR to nuklear with this change?

components/misc/misc.h
```diff
@@ -24,6 +24,65 @@ namespace Misc
       Point operator- (const Point &v) const { return {x - v.x, y - v.y}; }
       Point operator* (double c) const { return {static_cast<int32_t> (x * c), static_cast<int32_t> (y * c)}; }
     };
+
+    namespace detail
```
wheybags: Could you explain what all this is for? And maybe add a comment with explanation?

resources/gui/bottommenu.rml
wheybags: Actually, we should just delete this whole directory now

Predelnik: Well I was removing all the rml/py changes which were having merge conflicts, so some slipped. But yes it probably should be removed and used only for reference.

components/misc/misc.h
Predelnik: There is a little comment at `overload` function. Basically this is the function which makes usage of `variant` classes bearable. If you have it you can write something like:
`overload ([](int){}, [](std::string){})` and use it as a variant visitor with additional benefits like lambda captures etc. Writing separate class for visitor is often cumbersome.

extern/nuklear/nuklear.h
Predelnik: Yeah I was going to ask about it in the issue, since maybe this should actually be behavior 
 of `nk_widget_has_mouse_click_down` function but since my previous 2 issues still unanswered I was hesitating.

Predelnik:
I will figure out build issues tomorrow.

wheybags:
Sweet, looks really nice

PR #289, opened 2017-08-27
Update boost version on travis to the one used for windows build.

Predelnik:


wheybags:
You'll need to add a ppa with that version of boost.

Predelnik:
It seems like there's no such ppa at least for precise. I'm closing this for now, would think of another approach.

PR #290, opened 2017-09-02
Implement item highlighting on hover and checkerboard pattern for two-handed weapons.

Predelnik:
Sadly one more fix was added to nuklear but hopefully temporarily. I removed one of the previous ones though.
In general I think it maybe useful to use the same fragment shader for gui as for sprites since highlighting logic is the same but it's not the biggest code duplication ever so maybe that's fine. Overall I think using shaders for this kind of stuff makes sense.

### DIFF
--- a/apps/freeablo/fagui/guimanager.cpp
+++ b/apps/freeablo/fagui/guimanager.cpp
@@ -66,7 +66,7 @@ namespace FAGui
         nk_style_item tmpBg = ctx->style.window.fixed_background;
         struct nk_vec2 tmpPadding = ctx->style.window.padding;
 
-        ctx->active = ctx->current; // yet another hack for background windows
+        ctx->active = ctx->current;
         ctx->style.window.fixed_background = nk_style_item_image(background);
         ctx->style.window.padding = nk_vec2(0, 0);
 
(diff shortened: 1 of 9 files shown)


wheybags:
Nice, thanks!

wheybags:
note: I get this warning on my machine (gcc 6.3):
```
/home/wheybags/freeablo/apps/freeablo/faworld/inventory.cpp: In member function ‘FAWorld::Inventory::ExchangeResult FAWorld::Inventory::needsToBeExchanged(const FAWorld::Item&, const FAWorld::EquipTarget&) const’:
/home/wheybags/freeablo/apps/freeablo/faworld/inventory.cpp:140:19: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
                   for (auto i = target.posX; i < target.posX + item.getInvSize().first; ++i)
                   ^~~
/home/wheybags/freeablo/apps/freeablo/faworld/inventory.cpp:147:21: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘for’
                     return result;
                     ^~~~~~
```

Predelnik:
Yeah that's just indent issue I think I'll just reformat the whole file.

PR #292, opened 2017-09-09
Implement basic support for rendering text using smaltext.cel.

Predelnik:
Also I applied it for item description (as demo)


wheybags:
Nice, thanks!

PR #293, opened 2017-09-17
Item dropping and pick-up (mostly)

Predelnik:
Also this PR contains actor targeting/highlighting and display their description.
Plus a bit of changes on how mouse click is handled to be closer to how it's processed in Diablo.
Some slightly changes to pathfinding was made also to make it closer to original.
I haven't touched netcode because I don't know how to test it but I think that all that was working there before should continue to work.


pengyz:
So good a news it is !

wheybags:
Nice! This is starting to resemble a playable game :3

wheybags:
As for netcode, don't worry about it, I think I'm probably gonna do a fairly major rewrite soon :s

PR #298, opened 2017-10-14
Make use of animation sequences specified for some towners.

Predelnik:
So most of the towners have hardcoded animation sequences (sequences of frames) in game because sometimes their animation takes complex repeating pattern. This patch should extract them from diablo.exe and apply to towners which use it.

The other change I made is to make `animationPlayer` accept duration of single frame rather than duration of the whole animation. I think that originally Diablo has the similar timing in regards to animation (each frames takes roughly the same time). Since some of these sequences rather long it would be really bothersome to specify length of each animation manually. Also attack duration seems to look fine this way without previously applied fix, so I think it should be correct.

### DIFF
--- a/apps/freeablo/farender/animationplayer.cpp
+++ b/apps/freeablo/farender/animationplayer.cpp
@@ -7,36 +7,54 @@ namespace FARender
         if (mCurrentAnim == nullptr)
             return std::make_pair <FARender::FASpriteGroup*, int32_t >(nullptr, 0);
 
+        int32_t currentFrame;
         float progress = ((float)mTicksSinceAnimStarted) / ((float)mPlayingAnimDuration);
-        int32_t currentFrame = progress * mCurrentAnim->getAnimLength();
 
-        if (currentFrame >= (int32_t)mCurrentAnim->getAnimLength())
+        if (mPlayingAnimType != AnimationType::BySequence)
         {
-            if (mPlayingAnimType == AnimationType::Once)
+            currentFrame = progress;
+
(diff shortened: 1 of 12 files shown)


apps/freeablo/farender/animationplayer.cpp
```diff
@@ -7,38 +7,58 @@ namespace FARender
...
-            if (mPlayingAnimType == AnimationType::Once)
+        if (mPlayingAnimType != AnimationType::BySequence)
             {
-                playAnimation(nullptr, 0, AnimationType::Looped);
-                return getCurrentFrame();
-            }
-            else if (mPlayingAnimType == AnimationType::FreezeAtEnd)
-            {
-                currentFrame = mCurrentAnim->getAnimLength() - 1;
+                currentFrame = progress;
```
wheybags: dodgy indentation in this file

apps/freeablo/faworld/actoranimationmanager.cpp
```diff
@@ -50,6 +53,16 @@ namespace FAWorld
 
         // loop idle animation if we're not doing anything else
         if (sprite == nullptr)
-            playAnimation(AnimState::idle, FARender::AnimationPlayer::AnimationType::Looped);
+            {
```
wheybags: and here

Predelnik:
Fixed both cases. I have to deal with GNU coding style at work and sometimes it shows :)

wheybags:
:D np, thanks for the PR

PR #301, opened 2017-10-20
Allow building on OpenBSD.

ibara:
Hi --

This pull request fixes the freeablo build on OpenBSD.

Thanks!

### DIFF
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -75,7 +75,7 @@ if (WIN32)
     set(SRCS ${SRCS} nativefiledialog/src/nfd_win.cpp)
 endif()
 
-if ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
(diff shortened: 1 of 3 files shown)


extern/CMakeLists.txt
```diff
@@ -79,6 +79,10 @@ if ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
     set(SRCS ${SRCS} nativefiledialog/src/nfd_zenity.c)
 endif()
 
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "OpenBSD")
```
wheybags: Could you combine this with the linux block using AND instead of copying it?

ibara: Yeah, that's fine. I originally had it that way but then separated it because when the other BSDs come around it would become an unwieldy long long.

And yes, it does work.

ibara:
Done.

wheybags:
Thanks!

PR #306, opened 2017-11-08
Load path to dropping animation from exe instead of hardcoding.

Predelnik:
While doing that I also supported reading offsets in exe from .ini in '0x123ABCD' form and converted all the current offsets. This is just a stylistic change but make things a bit more readable in my opinion.

The change itself is also not hard - there is a list of possible dropping animations in .exe and mapping from index of inventory item picture to index in that list.
The same could be done with sounds in the future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,13 +73,14 @@ if(UNIX)
 endif()
 
 if(MSVC)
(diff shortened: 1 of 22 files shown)


wheybags:
Thanks again!
Hex is much better yeah

PR #307, opened 2017-11-13
Disable opengl es support (for now), force NVidia/AMD gpu if present.

### LINKED ISSUE #297: Lines between textures.
image
When I run freeablo on my laptop, there is this issue with rendering textures. (note green lines between tiles). I don't have much experience with opengl to figure out what may cause this, especially since it's obviously inconsistent on different GPUs/systems. I wonder if anyone has any clue what could cause this?

My laptop is ROG GL552VX running win10.

Predelnik:
Fixes #297.

### DIFF
--- a/components/render/sdl2backend.cpp
+++ b/components/render/sdl2backend.cpp
@@ -38,6 +38,13 @@
 #define MAX_VERTEX_MEMORY 512 * 1024
 #define MAX_ELEMENT_MEMORY 128 * 1024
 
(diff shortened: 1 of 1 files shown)


wheybags:
I'd prefer not to drop es support, because my laptop can only compile es shaders :(

Predelnik:
Ok let's just force AMD/Nvidia for now then. I will apply the changes soon.

PR #308, opened 2017-11-13
Optimization: Remove allocation/push backs during texture decoding.

Predelnik:
Reduces startup time on windows with MSVC's ITERATOR_DEBUG_LEVEL=2.

By the way assert which I commented fails somewhere. I didn't check where yet but looks like some image might have incorrect size.


wheybags:
Thanks!

PR #309, opened 2017-11-13
Add dummy dialogs for town citizens.

### LINKED ISSUE #273: re-add dialogue menus
Before we switched gui library from librocket to nuklear, we had dialog windows when you clicked on npcs

https://github.com/wheybags/freeablo/commit/abe20a463196964291f7121739c419f5a9afe518 is the last commit before merging the gui replacement, you can see this branch for an example.

Predelnik:
The following basic things yet need to be done:
- ~~Dialogs currently do not block clicks on the rest of the screen.~~
- ~~Mouse interaction~~
- ~~Escape button for now does not work because it should not call main menu in case of dialog (which is sort of similar to first item)~~
- Long press on arrow button should trigger fast scrolling between items.

But more or less it should fix #273


Predelnik:
So since PR is not yet merged I added fixes for everything I mentioned in the comment except long press. To support long press perhaps possibly I would have to refrain from using nuklear input but I don't think it's a significant feature anyway.

apps/freeablo/fagui/dialogmanager.cpp
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    DialogManager::DialogManager(GuiManager& gui_manager): mGuiManager(gui_manager)
+    {
+    }
+
+    void DialogManager::talkOgden()
+    {
+        DialogData d;
+        d.header({"Welcome to the", "Rising Sun"});
```
wheybags: I'd prefer not to have this text in here. If we're to have placeholder text, just some lorem ipsum would be better

Predelnik: Sorry, I don't quite get that. You mean I should extract strings from exe? But before that it was in writtine in .rml all the same, what's the difference?

wheybags: Yeah, I didn't know it was written in the rml before :s
IMO it's better to have placeholder text so we know it's not implemented, than to have the real text and forget it's hardcoded instead of loaded properly.

Predelnik:
I've done it.
To not create a ton of variables I just added a dictionary of strings for each character with (I hope) understandable keys. Variable for npcs in ini that are prefixed with `talk#` are processed automatically and added to their corresponding dictionaries.

wheybags:
Nice, thanks!

PR #310, opened 2017-11-13
Add parallel option to appveyor.

Predelnik:
Should speed up appveyor builds.


Predelnik:
Well 7 minutes against 9, not a huge win but already something. Probably should turn on /MP next.

wheybags:
I've been thinking we could maybe try https://github.com/frerich/clcache, and ccache for linux

PR #311, opened 2017-11-29
Update nuklear/ remove all custom fixes.

Predelnik:
Since all nuklear's bugs of interest have been succesfully fixed I'm reverting it to its master version and adding one more workaround (nastier version of which I removed in my previous commits)


wheybags:
>  reverting it to its master version

woohoo!
> adding one more workaround

:(

Any plans to upstream the last fix?

Predelnik:
> Any plans to upstream the last fix?
Well that's a bit of different matter, it's workaround external to nuklear.h. I've been thinking of reporting it but in its essence it is a matter of opinion should hover work on inactive windows or not, in Windows it does but maybe it's different in other UIs. For now it's fine as it is I think.

PR #312, opened 2017-12-02
Re-adding main menu (first stage)

### LINKED ISSUE #272: Re-add main menu
Before we switched gui library from librocket to nuklear, we had a mian menu on game startup.

https://github.com/wheybags/freeablo/commit/abe20a463196964291f7121739c419f5a9afe518 is the last commit before merging the gui replacement, you can see this branch for an example.

See also discussion on the new gui/opengl PR: https://github.com/wheybags/freeablo/pull/250#commitcomment-22785174

Predelnik:
Known Issues (I may fix some of them even before merge):
* ~~No mouse interaction~~
* Long press on arrow keys doesn't work (same as with dialogs)
* For now I added only initial main menu screen, I believe with librocket there was also character selection screen and maybe something more (?)
* ~~Letter width is a tad incorrect in pcx fonts for now, needs to be read from `.bin` files. (Also for smalltex.cel font it could be read from exe)~~

Once completed should fix #272

### DIFF
--- a/apps/freeablo/farender/fontinfo.h
+++ b/apps/freeablo/farender/fontinfo.h
@@ -2,6 +2,7 @@
 #include "fa_nuklear.h"
 
 #include <array>
+#include <SDL_surface.h>
 
 namespace Cel {
     class CelDecoder;
@@ -9,10 +10,10 @@ namespace Cel {
 
 namespace FARender
 {
-class FontInfo
+class CelFontInfo
 {
 private:
-    using self = FontInfo;
+    using self = CelFontInfo;
 public:
     nk_user_font nkFont;
 private:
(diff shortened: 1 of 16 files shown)


apps/freeablo/farender/fontinfo.h
```diff
@@ -2,25 +2,46 @@
 #include "fa_nuklear.h"
 
 #include <array>
+#include <SDL_surface.h>
```
wheybags: I'd prefer to avoid referencing SDL directly outside of the Render component. Could whatever this is being sued for be wrapped in some way in there?

Predelnik: Funny enough I think the easiest way to fix it also happens to be the most proper one. I just need to read correct character widths from `.bin` files, I used SDL surface to calculate them.

Predelnik: Fixed. Plus added mouse interactions.

wheybags:
Thanks!

PR #315, opened 2017-12-10
Make pause menu diablo-like.

Predelnik:
Also support reading character sizes for small dialog font from exe.

### DIFF
--- a/apps/celview/main.cpp
+++ b/apps/celview/main.cpp
@@ -2,6 +2,7 @@
 //#include <QApplication>
 
 #include "../components/settings/settings.h"
+#include <boost/format.hpp>
 #include <chrono>
 #include <faio/fafileobject.h>
 #include <input/inputmanager.h>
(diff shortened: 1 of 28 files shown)


wheybags:
Couple of small issues: 
- Your mouse click suppression code is a little overzealous, If I open the pause menu and then exit it without clicking, just by pressing esc again, my next mouse click will be ignored.
- IIRC originally changing menu option would trigger a sound to play
- Hovering a menu item should switch the pentagrams to it

But, yeah, all minor so I'll merge now anyway. Thanks again!

Predelnik:
* Hmm, mouse suppression is bad yeah, and looks like I made the same mistake with the one for dialogs, hope I'll fix it soon.
* For now I completely ignored the sounds and also music for main menu, should probably add it since some of it was implemented in pre-nuklear era iirc.
* I don't think pentagrams switch on hover in original game. I understand from UI standpoint this may make sense, but I guess it's better to reproduce Diablo behavior

Thanks for the merge!

PR #316, opened 2017-12-11
Fix blocking next mouse press in case of exiting dialog/pause menu using keyboard.

Predelnik:
Still feels a bit like workaround but for now this will do.


wheybags:
Nice, thanks, sorry for the delay in merging.

PR #317, opened 2017-12-16
Implement class selection menu.

Predelnik:
As a minor bonus I restored correct behavior for changing character sprite based on currently equipped weapon/armor.
Unimplemented for now - double click for selecting character class. It will probably require refactoring out input using nuklear just as arrow key's autorepeat.


wheybags:
Nice, thanks!

PR #331, opened 2017-12-25
now uses #pragma instead of header guard.

Hydroque:
Only `#pragma once` and spacing issues in the respective header files were changed. Nothing else. Commands used `git commit *.h -m "now uses #pragma instead of header guard."` `git push`.

### DIFF
--- a/apps/freeablo/engine/engineinputmanager.h
+++ b/apps/freeablo/engine/engineinputmanager.h
@@ -1,5 +1,5 @@
-#ifndef FA_INPUT_H
-#define FA_INPUT_H
+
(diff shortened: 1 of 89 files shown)


Hydroque:
travis-ci needs build 2 re-run, as I ran the auto-format patch and it actually worked this time.

Hydroque:
Ready for merging :)

wheybags:
Thanks!
For future reference, it's preferable to squash/rebase a or like this into one commit, since it's just small patches on the original commit, but it's fine, I can just use the github squash merge button :)

PR #328, opened 2017-12-22
now using #pragma instead of header guards, fixed spacing formating, too, on all H's @Hydroque

Hydroque:
Check if compiles successfully before using, just in case I overwritten something that depended on a `#define` Just simple spacing fixes so things aren't crowded together with everything .h converted to `#pragma`. I didn't touch anything from `freeablo/extern` see #325

PR #332, opened 2017-12-25
Patched - now using #pragma instead of header guards, fixed spacing formating, too, on all H's @Hydroque

Lectem:
This is #328 with the clang-format patch applied


wheybags:
Original or was fixed and merged, so closing this

PR #335, opened 2017-12-29
Refactor item related enums and move them to separate file.

Predelnik:
This should be mostly no-op PR to ease future working with items. Some unused effects calculation was removed for simplicity.
However I also attached some minor changes/bugfixes:
1. Hotkeys for opening panels.
2. Fix double creation of dialogs.
3. Support ability to heal for Pepin.


apps/celview/CMakeLists.txt
```diff
@@ -2,5 +2,10 @@ add_executable(celview
     main.cpp
 )
 
-target_link_libraries(celview NuklearMisc SDL2::SDL2main)
+target_link_libraries(celview NuklearMisc)
+if (MINGW)
+target_link_libraries(celview -lmingw32 SDL2::SDL2main)
```
wheybags: Is this only needed for celview?

apps/freeablo/faworld/actor.cpp
```diff
@@ -146,6 +146,8 @@ namespace FAWorld
         }
     }
 
+    void Actor::heal() { mStats.mHp = 100; }
```
wheybags: I'd do mStats.mHp = mStats.mHp.max;

apps/celview/CMakeLists.txt
Predelnik: Also for launcher. Everything else doesn't include SDL2main

apps/freeablo/faworld/actor.cpp
Predelnik: Ok

apps/celview/CMakeLists.txt
Predelnik: Hopefully it will be resolved upstream eventually then fix can be removed.

PR #340, opened 2018-01-02
Fix interaction between pause menu and inventory panel. (#338)

### LINKED ISSUE #338: Pause menu is buggy when you have inventory open
- cannot click menu entries
- If I have an item in my cursor, it stays as my cursor

Predelnik:
Should fix #338
Also disables hover on items during pause. Note that if something is hovered before pressing pause description message and highlight should stay until pause is finished, that's true to original game.

### DIFF
--- a/apps/freeablo/engine/enginemain.cpp
+++ b/apps/freeablo/engine/enginemain.cpp
@@ -232,6 +232,8 @@ namespace Engine
 
     const DiabloExe::DiabloExe& EngineMain::exe() const { return *mExe; }
 
+    bool EngineMain::isPaused() const { return mPaused; }
+
     void EngineMain::stop() { mDone = true; }
 
     void EngineMain::togglePause() { mPaused = !mPaused; }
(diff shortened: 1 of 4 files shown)


Hydroque:
I wonder if you can just save the current frame to a frame buffer while paused and cast new stuff over the top.

wheybags:
That might be the easiest thing actually

Predelnik:
That depends if you want to support change of resolution during pause or not.
And it actually completely opposite for multiplayer - in original game pause menu doesn't block anything in multiplayer. So you would have to support 2 pause menu workflows in this case.

wheybags:
Yeah, both fair points actually

Hydroque:
Changing of resolution only requires you to render, not do game logic. Supporting 2 pause menu workflows is as easy as skipping and not skipping the game logic section. It's definitely not daunting.

Unless you guys mixed game logic with rendering...

PR #343, opened 2018-01-06
Basic support for selling of items + some minor changes.

Predelnik:
1. I've removed `ItemManager` which was obsolete mostly anyway and created `ItemFactory` which only job should be to create items and randomize some of its properties. Basic items now created using id which is simple enum.
2. Implemented two types of description for basic items - for showing in description box while hovering in inventory and for dialogs. Hovering over dropped item displays just its name.
3. Implemented basic support for dialogs having more items than fit on the screen and other features required for selling/buying items.
4. As a proof of concept implemented selling items at Griswold's.

Missing for now:
- ~~Scrollbar~~.
- Filters by item type specific to merchant. After implmenting this selling of items to other merchants could also be added.
- Confirmation Dialog.


Predelnik:
So since I've broken serialization of items a bit I've decided to redo their implementation to simplify it. So technically all what is needed to be serialized for items is:
* base id
* things that are randomized (durability, armor class, charges)
* prefix/suffix id and its randomized value (1 or 2)
* unique id for unique items

The rest could be easily restored using `DiabloExe`. So for now while affixes are not supported only first 2 things could be serialized. However it creates problem that `Item` should have access to `DiabloExe` to return its charcteristics. It could be solved in 3 ways:
1. global access to `DiabloExe`
1. save pointer to `DiabloExe` inside `Item` during construction
1. pass `DiabloExe` to function calls.

For now I've solved it using second option, later it could be redone if deemed necessary.

Note that actors currently suffer from the same problem - they load/save some stuff which cannot be changed at runtime and could be easily restored from `DiabloExe`.

apps/freeablo/faworld/item.h
```diff
@@ -19,121 +19,116 @@ namespace FARender
...
-        uint8_t mSizeX = 0;
-        uint8_t mSizeY = 0;
-
-        uint8_t mCornerX = 0;
-        uint8_t mCornerY = 0;
+        const DiabloExe::DiabloExe* mExe = nullptr; // since we want to restore base info from exe data
 
         friend class Inventory;
-
-        // variables that used to be in this class, loaded form the exe, but ommented out for now sine we don't use them.
```
wheybags: I wouldn't get rid of these

apps/freeablo/faworld/playerfactory.cpp
```diff
@@ -24,61 +25,51 @@ namespace FAWorld
...
         {
-            item = itemManager.getItemByName("Potion of Healing");
-            player->mInventory.putItemUnsafe(item, FAWorld::MakeEquipTarget<EquipTargetType::belt>(i));
+            player->mInventory.putItemUnsafe(mItemFactory.generateBaseItem(ItemId::potionOfHealing), MakeEquipTarget<EquipTargetType::belt>(i));
         }
 
         player->setSpriteClass("warrior");
         player->mAnimation.setAnimation(AnimState::idle, FARender::Renderer::get()->loadImage("plrgfx/warrior/wld/wldst.cl2"));
         player->mAnimation.setAnimation(AnimState::walk, FARender::Renderer::get()->loadImage("plrgfx/warrior/wld/wldwl.cl2"));
+        // loadTestingKit (player);
```
wheybags: ?

components/diabloexe/baseitem.h
```diff
@@ -12,39 +12,42 @@ namespace DiabloExe
...
-        uint8_t reqDex;
-        uint8_t reqVit;
-
-        uint32_t specialEffect;
-        uint32_t magicCode;
-        uint32_t spellCode;
-        uint32_t useOnce;
-        uint32_t price1;
-        uint32_t price2;
+        uint32_t minArmorClass; // LONG LIVE OUR GLORIOUS QUEEN
```
wheybags: Comment no longer maxes sense when you change the spelling to US English :p

apps/freeablo/faworld/playerfactory.cpp
Predelnik: Well that's stuff I use for testing items, but maybe it's better to remove than comment it of course.

apps/freeablo/faworld/item.h
Predelnik: Well I removed them because I replaced most of them with functions which return stuff from `DiabloExe` for now since without affixes almost anything is constant anyway. With affixes some sort of state would be required but I think it'd be better to separate it from `Item` itself and store it as some other class inside it because for example it does not need to be serialized. But anyway I think it's something which needs to be thought through a bit later.

wheybags: Ok, that makes sense

apps/freeablo/fagui/dialogmanager.cpp
```diff
@@ -1,58 +1,81 @@
...
+    DialogLineData::DialogLineData() {}
+
     DialogLineData& DialogLineData::setAction(std::function<void()> actionArg)
     {
         action = actionArg;
         return *this;
     }
 
-    DialogLineData& DialogData::text_lines(const std::vector<std::string>& texts, TextColor color, bool alignCenter)
+    DialogLineData& DialogLineData::setNumber(int number)
```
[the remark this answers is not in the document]
Predelnik: Yeah I replace the `int` in this file but probably I create a lot of `int`s anyway, it's not super easy to keep track of this stuff :(

components/diabloexe/baseitem.h
Predelnik: Fixed :)

Predelnik:
I fixed bug of missing gold, which by the way uncovered that proper gold merging and pick-up needs to be implemented.
Can't reproduce freezing for now. Maybe some call stack would help.

wheybags:
Here's the stack of the game thread if I pause it:
image

apps/freeablo/faworld/item.h
wheybags: It's a matter of choice really, whether to read original files on the fly or convert at first launch. I opeted for read on the fly, and don't plan to change that for now unless there's a compelling reason down the line.

Predelnik:
Thanks! Looks like I was able to reproduce it finally in form of a crash. Hopefully I fixed this in an update. Would be great if you check it.

apps/freeablo/faworld/item.h
Hydroque: Modification of resources for modders.

wheybags:
Yup, working for me now

Predelnik:
Ok, thanks! I'm merging then.

apps/freeablo/faworld/item.h
wheybags: Sure, but mods aren't a priority until the game works. And besides, we can already load pngs and the likes anyway, we don't need to convert all the cels to pngs for that

PR #344, opened 2018-01-06
Test speed of ninja build on appveyor.

Predelnik:


Predelnik:
Ok so it seems using boost with ninja is currently not possible
https://github.com/ruslo/hunter/issues/1223
Maybe I'll look into that. For now - not closing this PR

Predelnik:
I found workaround how to make build with ninja pass using double run of cmake. However I never experience hunter cache hit on appveyor that way for some reason.
Anyway by my research Ninja build will get fixed after Hunter will be using newer boost for MSVC but it was unluckily failing with versions 1.65, 1.66.
Anyway, for now closing.

PR #345, opened 2018-01-14
Complete selling implementation + some small fixes.

Predelnik:
1. Completed selling filters and confirmation dialog.
2. Walking through stairs is probably temporary solution and technically they should be more than one tile in size, but I think that it's still important for them to work at least this way for now.
3. Fix for character look not updating to dungeon sprites when in dungeon.

### DIFF
--- a/apps/freeablo/fagui/dialogmanager.cpp
+++ b/apps/freeablo/fagui/dialogmanager.cpp
@@ -30,12 +30,20 @@ namespace FAGui
         return *this;
     }
 
-    DialogLineData& DialogData::textLines(const std::vector<std::string>& texts, TextColor color, bool alignCenter)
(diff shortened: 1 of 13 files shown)


stavrossk:
Hi, is Freeablo playable at it's current state?

Thanks,
Stavros.

wheybags:
Nope, not really, but getting close

stavrossk:
@wheybags Thanks, can't wait to play Diablo on my modern computer!

Hydroque:
@stavrossk There are CD patches available so that you can play and you'll need a fixed directx dll iirc. Although, for freeablo, you'd already have to meet the requirement as it uses its copyrighted resources, but doesn't distribute.

PR #348, opened 2018-02-11
Make equipped weapon affect damage in some way.

### LINKED ISSUE #327: proper direction type
We just use an into for direction, where the values 0-7 encode the compass directions.
This is what enums are for :p

Predelnik:
Changes:
1. Refactor `Direction` to `enum class` and simplify functions working with it, should fix #327
2. Move all the `Random` stuff to namespace random and to `Misc` component
3. Make melee attack apply damage not at the start but on some animation frame, currently the middle of animation (placeholder)
4. Implement shift+attack
5. Add some incomplete formulas for player attack, full of placeholders for now.
6. Implement some basic mechanism for summing up stats from equipped items, for now only applies to damage but should be easily expandable.

So overall it should not change anything much but rather serve as a groundwork for the future changes. I just wanted a mechanism for items to have at least some effect on the game before starting to implement item generation.


apps/freeablo/faworld/itembonus.h
```diff
@@ -0,0 +1,33 @@
...
+}
+
+namespace FAWorld
+{
+    /// ItemBonus is a class for storing addable properties of items which later applied to a player
+    /// It was decided to not use enum indexed vector for it because some of stats should be summed as a bitmask
+    /// In case of performance problem sum of ItemBonus'es can later be cached and updated only in case if equipment is changed
+    class ItemBonus
+    {
+        using selfType = ItemBonus;
```
wheybags: why not just write ItemBonus?

components/CMakeLists.txt
```diff
@@ -85,7 +89,7 @@ add_library(DiabloExe
     diabloexe/characterstats.cpp
     diabloexe/affix.cpp
     diabloexe/affix.h)
-target_link_libraries(DiabloExe Misc FAIO ${HUNTER_BOST_LIBS})
+target_link_libraries(DiabloExe Misc FAIO ${HUNTER_BOOST_LIBS})
```
wheybags: lol woops

components/misc/misc.cpp
```diff
@@ -1,106 +1 @@
 #include "misc.h"
-
```
wheybags: maybe just remove this file?

apps/freeablo/faworld/itembonus.h
Predelnik: Well it's matter of preference, I prefer to avoid repeating type's name too much, so it would be easier to move function to another class for example. I also used to name this typedef just `self` which makes it much shorter but I've seen some people complaining about it so I've tried adding `type` in this case.

components/misc/misc.cpp
Predelnik: Ok

PR #351, opened 2018-02-25
Fixes for handling gold in inventory and gold split dialog.

Predelnik:
Also I fixed reading strings with spaces from ini which was broken since I've added reading of hex values and that was really bad for paths.


resources/hotkeys.ini
```diff
@@ -67,4 +67,59 @@ alt=0
 key=98
 shift=0
 ctrl=0
-alt=0
\ No newline at end of file
+alt=0
+[Digit0]
```
wheybags: We don't really need to have these in a config file do we? They don't make sense to have a somethung configurable.

Predelnik: Yeah but that would require some new mechanism for passing pressed keys, I wanted to postpone this until things like inputting player name would be needed :)

Well guess it wouldn't be hard to implement that now

PR #355, opened 2018-02-26
inventory fixes

wheybags:


apps/freeablo/fagui/guimanager.cpp
```diff
@@ -430,8 +430,8 @@ namespace FAGui
             nk_button_label_styled(ctx, &dummyStyle, "");
             if (nk_widget_is_mouse_click_down_inactive(ctx, NK_BUTTON_LEFT))
             {
-                inv.inventoryMouseLeftButtonDown(Misc::Point{int32_t((ctx->input.mouse.pos.x - invTopLeft.x - ctx->current->bounds.x) / cellSize),
```
Predelnik: Does this really help though? We're still pasing exact cell not based on item size. But for certain items its center is in the middle of the cell and for others it's in corner of one cell or in the middle of cell side, so it should condier size of the item before rounding.

wheybags: Hmm yeah, this does break it for other items. I'll do a "proper this time I swear" fix later :p

apps/freeablo/engine/enginemain.cpp
```diff
@@ -179,7 +179,10 @@ namespace Engine
                     state->mCursorEmpty = false;
                 else
                     state->mCursorEmpty = true;
-                state->mCursorFrame = FAGui::cursorFrame;
+                if (mWorld->getCurrentPlayer())
```
wheybags: @Predelnik This is the wip inv refactor, it's almost done but I didn't quite finish it today

apps/freeablo/faworld/inventory.cpp
```diff
@@ -19,6 +19,190 @@ using namespace boost::container;
...
+        {
+            for (uint8_t x = 0; x < mInventoryBox.width(); x++)
+            {
+                mInventoryBox.get(x, y).mInvX = x;
+                mInventoryBox.get(x, y).mInvY = y;
+            }
+        }
+    }
+
+    bool Inv::autoPlaceItem(Item& item, PlacementCheckOrder order)
```
Predelnik: Well here's definitely going to be a conflict. I had to separate autoplacement function so it would return location first and I could check that item fits before buying and then place it to appropriate location.

PR #357, opened 2018-03-07
Buying basic items + some player stats display

Predelnik:
1. Player stats are experimental but I do think they should be separated from actor stats because players have a lot of unique properties.
2. Buying basic items from Griswold is the easiest kind of item generation. Current limitations:
* items not saved/loaded.
* not regenerated on reentering town.
* `ilvl` is fixed for now.

### DIFF
--- a/apps/freeablo/faworld/inventory.cpp
+++ b/apps/freeablo/faworld/inventory.cpp
@@ -113,7 +113,7 @@ namespace FAWorld
         {ItemEquipType::head, {EquipTargetType::head}},
     };
 
-    bool Inventory::isFit(const Item& item, const EquipTarget& target) const
+    bool Inventory::couldBePlacedToInventory(const Item& item, const EquipTarget& target) const
     {
         switch (target.type)
         {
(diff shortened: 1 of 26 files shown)


wheybags:
merged into #355

apps/freeablo/faworld/inventory.cpp
```diff
@@ -316,6 +316,34 @@ namespace FAWorld
...
+                    continue;
+                auto toTake = std::min(quantity, item.mCount);
+                item.mCount -= toTake;
+                quantity -= toTake;
+                if (item.mCount == 0)
+                    item = {};
+                if (quantity == 0)
+                    return;
+            }
+        release_assert("Not enough gold");
```
wheybags: Side note, but this doesn't work as intended, as release_assert expects a boolean, it will cast the const char* to bool, and get true, so this will never fail. Can be worked aroudn by `release_assert(false && "message")`, but I've also added a  `message_and_abort` macro for thsi case. I also added `invalid_enum` for invalid enum values.

Predelnik: Will consider in the future. Thanks!

Issue #358, opened 2018-03-11
Issues with inventory after refactoring


Predelnik:
1. Removing hotspot change was in my opinion wrong decision. Technically as I mentioned in comment this behavior is not the same as in game. But emulating it is pretty hard (and requires forced cursor movement which may not be a good thing). Current problems due to this:
If top-left point is outside of inventory or slot item will never be placed which is not true in original game and is done if center is inside inventory or slot:
![image](https://user-images.githubusercontent.com/2133957/37252531-74cf498a-2533-11e8-9a02-63651eef0888.png)
Huge unintuitive movement of cursor when taking item especially clicking on bottom-right corner of the item:
![image](https://user-images.githubusercontent.com/2133957/37252549-98353d76-2533-11e8-9911-7c98d79350d5.png)
Clicking leads to:
![image](https://user-images.githubusercontent.com/2133957/37252553-a61c4d26-2533-11e8-9d44-fcc5705bb9d6.png)

This behavior is far from unnoticeable in my opinion.

2. Now items could not be placed in the belt at all if taken from it.
3. The same problem with placing buckler as before:
![image](https://user-images.githubusercontent.com/2133957/37252576-25fd7100-2534-11e8-9d09-f0b9d5dcc8a4.png)

4. Warrior starts with shield in inventory and not in hand.

wheybags:
1. Agreed, but my plan was to hilight the squares under the item which imo will make it much more intuitive. I was goign to do this in the same pr but I decided to leave it as it was getting quite huge already.
2. hmm, not sure what happened there
4. ah yes, I forgot to fix that, should be a one line change

Predelnik:
1. I don't get how squares would help with issues I mentioned if item as cursor is tied to top-left corner - if  you click above inventory it will still not work and without changing cursor position forcefully while clicking on bottom-right corner of the item, item cursor will still visually move huge distance if clicked.

wheybags:
Sure, it wont work but you won't even try as it it'll be obvious that it won't work. As for the item jumping when you click it, it's always going to happen to some degree unless we do super fancy adjustments for where you click it, which personally I don't think is worth it.

wheybags:
Anyway, I'm not saying it will work for sure, but I'd like to try it, as it keeps the code simpler. If it doesn't play well then we can do some fancy click position adjustment.
[...]

PR #359, opened 2018-03-11
Inv fixes 2

wheybags:


wheybags:
Fixes issues 2 and 4 from #358

PR #364, opened 2018-03-26
Update nuklear.h

Predelnik:
Fixes a possibility to hang with memory being allocated towards infinity.


wheybags:
Is this vanilla now, or is it still slightly different to upstream?

Predelnik:
It's vanilla from nuklear master as it was before. I didn't apply any patches for a while.

wheybags:
Ah nice

Issue #326, opened 2017-12-21
consistent Point type


wheybags:
We have a million different ways of representing a pair of int32_ts, sometimes we use std::pair<int32_t>, sometimes we pass separate x and y values, and there's a few places where we've even made structs with an x and y member.
It would be nice to create a single Point type (probably in the misc component), and just use that consistently. Note that the Position class cannot be used for this, as it represents more than just a single point (it represents two points side by side and an interpolation between them)

Hydroque:
Should it be a definite use of std::pair<T, V>, I mean it is well defined.

wheybags:
I'd prefer a dedicated class with an X and y member. That way we don't need
to write first and second all the time, and we can add member functions

Hydroque:
Correct, but standard pair is what we are looking for and then and if we need more we can abstract it. I mean we can scrap the std pair if you think it doesn't fit. It's an easy fix

Hydroque:
I am thinking of something like this, it's a header only

__pair.hpp__
```

#pragma once

namespace Misc
{
	template<typename T, typename V> class TPair
	{
	public:
		T x;
		V y;
		inline TPair(T x, V y)
		{
			this->x = x;
			this->y = y;
		}
	};
	
	template<typename T> class SPair : public TPair<T, T>
	{
	public:
		inline SPair(T x, T y)
			: TPair<T, T>(x, y)
		{
			
		}
	};
}

```

__Usage:__
```
#include <iostream>

#include "pair.hpp"

int main() {
	
	Misc::TPair<int32_t, int32_t> arr1(1, 2);
	Misc::SPair<int32_t> arr2(3, 4);
	
	std::cout << "first: " << arr1.x << ", " << arr1.y << std::endl;
	std::cout << "second: " << arr2.x << ", " << arr2.y << std::endl;
	
	return 0;
}
```

__Which should be the rough equivalent of https://godbolt.org/g/MNWMzz__
```
int32_t x1;
int32_t y1;
x1 = 1;
y1 = 2;

int32_t x2;
x2 = 3;
y2 = 4;
```

Predelnik:
What's the point of having templated pair when for point you want fixed type and definetily not different types for x/y? As for your `SPair` what's the cardinal difference with it and `std::array<int32_t, 2>`?
Simplest implementation of point ( + maybe the same with doubles):

    struct point
    {
        int32_t x;
        int32_t y;
    };

[...]

PR #375, opened 2018-07-21
added Misc::Point and in renderer.cpp Windows endings are replaced

akuskis:
This PR should be bound to ticket #326 
Misc::Point was added.
I would like to add get/set functions and rx/ry (for getting references), but in general looks like variables in project are not private, so I've left them public too. Later it can be improved.

### DIFF
--- a/components/misc/point.cpp
+++ b/components/misc/point.cpp
@@ -1,0 +1,12 @@
+#include "point.h"
+
+namespace Misc
+{
+
+    Point::Point() : Point(0, 0) {}
+
+    Point::Point(int32_t x, int32_t y) : x(x), y(y) {}
+
+    bool Point::operator<(const Point& other) const { return x < other.x || (x <= other.x && y < other.y); }
+
+} // namespace Misc
--- a/components/misc/point.h
+++ b/components/misc/point.h
@@ -1,0 +1,35 @@
+#pragma once
+
+#include <cinttypes>
+#include <functional>
+
+namespace Misc
+{
+    struct Point
+    {
+        int32_t x;
+        int32_t y;
+
+        Point();
+        Point(int32_t x, int32_t y);
+
+        Point operator+(const Point& v) const { return {x + v.x, y + v.y}; }
+        Point operator-(const Point& v) const { return {x - v.x, y - v.y}; }
+        Point operator*(double c) const { return {static_cast<int32_t>(x * c), static_cast<int32_t>(y * c)}; }
+        Point operator/(int c) const { return {x / c, y / c}; }
+        bool operator==(const Point& other) const { return x == other.x && y == other.y; }
+        bool operator!=(const Point& other) const { return !(*this == other); }
+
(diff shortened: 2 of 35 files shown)


akuskis:
In latest commit I've executed clang-format-6.0 on changed files.
But anyway, only this check doesn't pass...

components/misc/point.h
```diff
@@ -0,0 +1,35 @@
...
+    {
+        int32_t x;
+        int32_t y;
+
+        Point();
+        Point(int32_t x, int32_t y);
+
+        Point operator+(const Point& v) const { return {x + v.x, y + v.y}; }
+        Point operator-(const Point& v) const { return {x - v.x, y - v.y}; }
+        Point operator*(double c) const { return {static_cast<int32_t>(x * c), static_cast<int32_t>(y * c)}; }
```
wheybags: This should probably be a FixedPoint, not a double.
Not sure where it's used, but if we really need a double version (for rendering, we should never use it for the actual game logic), we can just do it inline.

wheybags:
Also, re: getters and setters, I normally only add a (const) getter, if I
find myself adding a setter for it later I just make it public, because
that';s what a getter/setter pair is anyway, just with nastier syntax.
And sure, it's a bit of a pain later if you want to actually have the
getters/setters run code, but that's what IDEs are for, and IMO it's better
to have nicer code now than sacrifice readability for a minor timesaving
that might well not even happen later.

akuskis:
Thank you for your comments!
I've found before, that there was used special patch and clang-format v5. I see your point about using the same version and patching... just afraid that from time to time that patch should be updated.
Yes, there are missed opportunities (as you said), because I'd like to get some feed back first. I'll take comments into account. Thank you for your review!

components/misc/point.h
akuskis: They are used in sdl2backend.cpp
I've marked this lines as "inline"

components/misc/point.cpp
```diff
@@ -0,0 +1,12 @@
+#include "point.h"
+
+namespace Misc
+{
+
+    Point::Point() : Point(0, 0) {}
+
+    Point::Point(int32_t x, int32_t y) : x(x), y(y) {}
+
+    bool Point::operator<(const Point& other) const { return x < other.x || (x <= other.x && y < other.y); }
```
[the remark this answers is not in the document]
akuskis: I usually hide all implementation in source and header is used only to explain interface. In this case I also do not set default values and keeps interface as simple as possible.
That is why I prefer to use single constructor as entry point `Point(int32_t x, int32_t y)` to create this object.
As code convention is different, I'll change code.

components/serial/loader.cpp
```diff
@@ -57,6 +65,12 @@ namespace Serial
 
     void Saver::save(const std::string& val) { mStream.write(val); }
 
+    void Saver::save(const Misc::Point& point)
```
wheybags: I would prefer not to do this here, and just add save() and load() methods to the Point class.
Otherwise, looks good to me :)

PR #381, opened 2018-08-19
Bug was fixed. Take correct item from array.

akuskis:
When it's required to get item in 2d array, `y` should be multiplied by `width` (not `height`).


akuskis:
Test for this behavior and explanation can be found around: https://github.com/wheybags/freeablo/pull/380#issuecomment-414082231

wheybags:
There is an implementation of Array2D in misc, this should probably be used instead.
Will merge for now anyway though

PR #382, opened 2018-08-19
better path finding

akuskis:
Take geometry into account (diagonal is longer then side in right-angled triangle).
`pathFind` does not change `goal` any more. Goal can be taken from result array.


akuskis:
Test for this behavior and explanation can be found around: #380

wheybags:
This code is the same changes as in the other PR right? If so, we can close this one and just merge them together.

PR #383, opened 2018-08-19
fix for Travis

akuskis:
It's very annoying when Travis fails because there is a problem in another part of code and for each PR I have to make the same changes to pass build.


wheybags:
Thanks, sorry for the delay!

PR #402, opened 2018-11-18
Server::mHost is null on Windows 10 without calling enet_address_set_…

### LINKED ISSUE #401: Server::mHost is null on Windows 10 without calling enet_address_set_host()
In `Engine::Server`, `enet_host_service()` was crashing because `mHost` was null on my Windows 10 machine after call to `enet_host_create()`.

```cpp
 void Server::update()
{
    ENetEvent event;

    while (enet_host_service(mHost, &event, 0))  // Crash here
    {
        ...
    }
}
```

```cpp
Server::Server(FAWorld::World& world, LocalInputHandler& localInputHandler) : mWorld(world), mLocalInputHandler(localInputHandler)
{
    mAddress.port = 6666;
    mAddress.host = ENET_HOST_ANY;
    mHost = enet_host_create(&mAddress, 32, 2, 0, 0);  // mHost will be null
}
```

Setting the address to the localhost ("127.0.0.1") with `enet_address_set_host()` solved the issue for me. Not sure if it was removed previously for some reason.

balintkiss501:
…host()

Fixes #401.

### DIFF
--- a/apps/freeablo/engine/enginemain.cpp
+++ b/apps/freeablo/engine/enginemain.cpp
@@ -142,6 +142,11 @@ namespace Engine
         mInputManager->registerMouseObserver(mLocalInputHandler.get());
         mInputManager->registerKeyboardObserver(mLocalInputHandler.get());
 
+        if (0 != enet_initialize())
+        {
+            std::cerr << "Unable to initialize networking library." << std::endl;
+            return;
+        }
         if (variables["client"].as<bool>())
         {
             mMultiplayer.reset(new Client(*mLocalInputHandler.get()));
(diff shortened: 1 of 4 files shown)


wheybags:
This sounds like a bug in enet. ENET_HOST_ANY should iirc bind all interfaces.
Could you make a minimal reproduction and report it upstream? For now I will merge this as a workaround.

wheybags:
Thanks for the PR!

balintkiss501:
> Could you make a minimal reproduction and report it upstream? For now I will merge this as a workaround.

I will confirm this with a simple C program and submit an issue to the enet GitHub Issues page. Might even go through the library with a debugger.

> Thanks for the PR!

Glad I could contribute!

PR #403, opened 2018-11-20
388 replace boost ini parser

### LINKED ISSUE #388: Replace boost ini parser
Boost ini parser is used on startup for settings + when loading CEL files.
It would be nice to replace this, as it uses exceptions internally, which completely breaks the ability to set a brakpoint on exceptions being thrown.
Using a single-header lib like https://github.com/jtilly/inih would probably be the best option.

balintkiss501:
Fixes #388


components/settings/settings.cpp
```diff
@@ -52,96 +73,224 @@ namespace Settings
...
+        return get<int64_t>(section, name, defaultValue);
+    }
+
+    template <> uint64_t Settings::get<uint64_t>(const std::string& section, const std::string& name, uint64_t defaultValue)
+    {
+        const std::string textValue = get<std::string>(section, name, std::to_string(defaultValue));
+        return std::stoull(textValue, nullptr, 0);
+    }
+
+#ifdef __APPLE__
```
wheybags: Should it be checking for clang, instead of apple?
I think a better solution might be to change https://github.com/balintkiss501/freeablo/blob/388_replace_boost_ini_parser/components/diabloexe/diabloexe.cpp#L150 to use uint34_t instead, so this overload isn't needed.

balintkiss501: > Should it be checking for clang, instead of apple?
Yes, `ifdef __clang__` would be definitely better.

But I'd be happier without this workaround and with the change in that `diabloexe.cpp` you mentioned. Looking back in the failed build log, this was the only function including that `auto` keyword that caused trouble.

balintkiss501:
First I thought about using `const` for the `defaultValue` and `value` parameters like this

```cpp
template <class T> T get(const std::string& section, const std::string& name, const T defaultValue = T());
template <class T> void set(const std::string& section, const std::string& name, const T value);
```

because it's my design philosophy that having const function parameters helps detecting bugs earlier while maintaining the function implementation. Not sure if you had other intentions by keeping those parameters as they are.

`const T& defaultValue = T()` would be great for `std::string` parameters, but utter waste for primitives like `bool` and `int`.

components/settings/settings.cpp
balintkiss501: Osx build failed again, @wheybags do you wish to replace occourences of `auto` with `size_t` in `diabloexe.cpp`?

wheybags: Can We just add an overload for size_t that forwards to uint64_t?

balintkiss501: I set Clang on my machine just for this and to me, it seems like

```cpp
#ifdef __clang__
template <> size_t Settings::get<size_t>(const std::string& section, const std::string& name, size_t defaultValue)

    return get<uint64_t>(section, name, defaultValue);
}
#endif
```

compiles on both GCC and Clang.

balintkiss501: I'm OK with your commit.

wheybags:
Thanks for the PR!

PR #407, opened 2018-12-06
Feat/393/esc to close dialogs

### LINKED ISSUE #393: Press Esc to close dialogs
Pressing esc should close whatever dialogs are open, until none are open, and only then open the pause menu

grantramsay:
Fixes #393, also fixed a few other similar minor inconsistencies.
I'm using a mac so can only compare against Belzebub, possibly some of this stuff is different from vanilla. Just let me know and I'll revert those commits


wheybags:
Looks good, thanks!

Issue #27, opened 2014-02-23
Video playback


wheybags:
ffmpeg is probably what we want here, but I'm no expert

konopka90:
http://dranger.com/ffmpeg/

Tutorial for SDL 1.2

wheybags:
I think we might actually consider using https://wiki.videolan.org/LibVLC/ instead.
As far as I understand, it's a lot higher level, and easier to use.
Plus, ffmpeg is an absolute pain to use, with it's two almost-compatible-but-not-quite forks

konopka90:
I'm not sure but LibVLC can't open SMK videos. Latest VLC can only play sound but no video :(

konopka90:
Can be useful https://github.com/pockethook/player

wheybags:
Hmm, yeah, vlc is just giving me audio too :s
That player crashes for me when I try to open an smk file though :s
```
[Arts-MacBook-Pro] ~/player [master] >
$ ./player ~/freeablo/build/logo.smk
[swscaler @ 0x7fb07a829600] bad dst image pointers
[swscaler @ 0x7fb07a829600] bad dst image pointers
[swscaler @ 0x7fb07a829600] bad dst image pointers
[swscaler @ 0x7fb07a829600] bad dst image pointers
[swscaler @ 0x7fb07a829600] bad dst image pointers
[swscaler @ 0x7fb07a829600] bad dst image pointers
Dis[swscaler @ 0x7fb07a829600] play error: SDL textbad dst image pointers
ure update
```

ffmpeg can definitely be made to work anyway, someone did a half baked implementation that got abandoned a while ago (actually, was that you? :p)

konopka90:
yup that was me ;p there was audio and video but without synchronization what is core of the problem. I'll try to debug this player, maybe it will work with some changes.

wheybags:
It worked fine on an mp4 file btw

konopka90:
Sounds good

konopka90:
Fail - it is possible to play SMK with small changes but this player doesn't support sound :-1:

Hydroque:
https://trac.videolan.org/vlc/ticket/17320
Check out this! It seem like master of vlc fully supports SMK files!

mewmew:
>  Check out this! It seem like master of VLC fully supports SMK files!

Wow, that's great! Never thought such an ancient format would be supported upstream :p

wheybags:
[...]

PR #410, opened 2018-12-14
Add video playback (libmpv)

grantramsay:
Fix for #27

Working video playback, only tested on a mac so far.
Just [install mpv](https://mpv.io/installation/)
Also has a stub video implementation that just skips videos if cmake can't find mpv.
Click esc to skip videos


wheybags:
This looks ok to me, but I get a link error (probably related to distrto packages)
```
/usr/bin/ld: //usr/lib/x86_64-linux-gnu/libopenmpt.so.0: undefined reference to `std::random_device::_M_getentropy() const@GLIBCXX_3.4.25'
```

Before something like this is merged, I'd prefer to switch to using a [hunter package](https://github.com/ruslo/hunter) 

It seems at the moment, there is no package for libmpv, so one would need to be added. Is this something you're interested in doing?

grantramsay:
That's annoying the build didn't work, mpv mention linux users are [better off to use 3rd party packages](https://mpv.io/installation/), not sure if that'd help.

I had a quick look at hunter, looks like quite a bit of effort to add libmpv...
i.e:
- would that involve also adding all libmpv dependancies to hunter? e.g. ffmpeg and it's many dependancies etc etc
- [libmpv can't be built with MSVC](https://github.com/mpv-player/mpv/blob/master/DOCS/compile-windows.md#linking-libmpv-with-msvc-programs), they recommend linking MSVC against an mpv library built with MSYS2 (I grabbed a windows VM, gave this a go and it can be made to work for this PR). Hunter seems more about building from source so they might not like including prebuilt libraries..?

wheybags:
> would that involve also adding all libmpv dependancies to hunter? e.g. ffmpeg and it's many dependancies etc etc
Unfortunately, yes :(
> libmpv can't be built with MSVC
From what I can see, they don't actually say it can't be, they just don't provide instructions for doing so.
In any case, we may be able to at some point.
I'm not sure what the best coiurse is here. Doing it properly is obviously a massive amount of work, but it would have a great benefit not only to this project, but also to hunter.
Maybe we could prepackage static libs for each platform for now (I would really like to stick to static libs, so we don't need any weird stuff when shipping, just one executable).

grantramsay:
"Maybe we could prepackage static libs for each platform for now"
Would that require a package for each linux distribution?
Be good if package managers were just kept up to date..

I might park this for a while until I have a better understanding of hunter and maybe look into that.
Anyone else feel free to carry this on etc

grantramsay:
Closing this for now

Issue #52, opened 2014-03-03
Figure out how arches work in the l1 tileset


wheybags:
see rnd2,3,5, for examples. Tileset does not seem to contain middles for the arches

d33tah:
> Attaching some email correspondence related to this issue, so the details are in the open.

Was this correspondence with the Diablo devs?

mewmew:
> Was this correspondence with the Diablo devs?

No, it was just an email conversation between wheybags and me during March last year, when he was still working on his dissertation. I decided to upload a copy of the correspondence to the relevant issue, as others may be interested in the details : )

wheybags:
This is in the game now

Issue #24, opened 2014-02-23
save/load


wheybags:
first implementation need only cover random levels + player position

konopka90:
#160

PR #411, opened 2018-12-20
Add the "special" sprites for arches / door frames

grantramsay:
Fix for #52
I'm not quite sure which arches #52 is referring to so haven't tested if that's working now..
Was more focussed on fixing the open door frames:
screen shot 2018-12-20 at 9 45 54 pm
screen shot 2018-12-20 at 9 47 53 pm


components/level/level.cpp
```diff
@@ -50,11 +61,20 @@ namespace Level
         Serial::ScopedCategorySaver cat("Level", saver);
 
         saver.save(mTilesetCelPath);
+        saver.save(mSpecialCelPath);
```
grantramsay: These save files seem a bit flaky, adding this new value likely breaks compatibility.
Probably should be using key value pairs at least?
#24

wheybags:
I'll be away fro christmas for a while, so I won't have time to look at this properly, but the screenshot looks great!
I've been wanting to get this in for ages now.
Thanks!

apps/freeablo/faworld/world.cpp
```diff
@@ -185,6 +185,21 @@ namespace FAWorld
                                    "levels/towndata/town.min",
                                    "levels/towndata/town.sol",
                                    "levels/towndata/town.cel",
+                                   "levels/towndata/towns.cel",
+                                   std::map<int32_t, int32_t>{{357, 1},
+                                                              {128, 5},
```
wheybags: This probably shouldn't be hardcoded, but for now I think it's fine since we'll probably have to change how levels are stored on disk anyway at some point (I'd like to eventually get rid of the min/til/dun hierarchy and just have "tiles").
Maybe for now you could add a comment to this and the other palce with hardcoded mappings, explaining what they are used for?

wheybags: Also, actually, what does this do in the town level?

grantramsay: Ok will add comments.
The town ones overlay on parts of trees, although they seems to make no difference to the final output...
Here's a screenshot with special cels highlighted:
screen shot 2019-01-05 at 8 03 56 pm
And here's all the special cel images in case anyone is interested:
levels-towndata-towns
levels-l1data-l1s
levels-l2data-l2s

wheybags:
Thanks!

PR #413, opened 2019-01-06
Add gif export to celview using a modified version of jo_gif.cpp

Exairnous:
Since a lot of CEL/CL2's are animations I have added an option to celview to save as an animated gif.  Transparency is supported and the resulting gifs open properly in gimp.  In all the tests I did there was never a problem with palette size -- thankfully diablo doesn't seem to use all 256 colors + transparency in each CEL/CL2.  The end filesize is not as small as it could be because a full 256 palette is generated for each frame, but since we're dealing with such low resolution I don't think it's a big problem.

I'm not sure if I should have modified cmake files to include the new library (jo_gif.cpp), but it compiles fine for me.  So, let me know if this needs any changes.

EDIT: I see travis is set to treat warnings as errors.  jo_gif.cpp started off with the warnings about the initializer list, but since it would compile on my system I just left it.  If you want I'll see if I can get rid of them.

EDIT2: I tried to fix the travis formatting issue with:

    curl https://termbin.com/t4hrn > /tmp/fa_format_diff.patch && git apply /tmp/fa_format_diff.patch

but first I got a certificate error and then when I connected with an insecure connection it couldn't find components/render/sdl2backend.cpp.  Any help would be appreciated.

### DIFF
--- a/apps/celview/main.cpp
+++ b/apps/celview/main.cpp
@@ -90,6 +90,17 @@ int main(int, char**)
                         }
                     }
 
+                    if (nk_button_label(ctx, "save as gif"))
+                    {
+                        nfdchar_t* outPath = NULL;
+                        nfdresult_t result = NFD_SaveDialog("gif", NULL, &outPath);
+                        if (result == NFD_OKAY)
+                        {
+                            Render::SpriteGroup::toGif(selectedImage, outPath);
+                            free(outPath);
+                        }
+                    }
+
                     auto msSinceLastFrame = std::chrono::duration_cast<std::chrono::milliseconds>(now - lastFrame).count();
                     if (animate && msSinceLastFrame > 100)
                     {
--- a/components/render/misc.h
+++ b/components/render/misc.h
@@ -25,6 +25,7 @@ namespace Render
         int32_t getHeight() const { return mHeight; }
 
         static void toPng(const std::string& celPath, const std::string& pngPath);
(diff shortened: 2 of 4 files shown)


Exairnous:
I looked into the brace initialization thing a bit more, and for the one on line 160 I think it could be fixed by replacing this:

    jo_gif_lzw_t state = {fp, 9};

with this:

    jo_gif_lzw_t state = {fp, 9, {}, *"", 0, 0, 0};


For the one on line 218, I think we should just leave it.  I'm not sure why travis is complaining about it, I don't get any warnings about it with gcc, and it seems to be pretty clear and standard c++.

Let me know what you think.

wheybags:
We already have headers for disabling warnings in extrenal files, I've added them now in the commit I just pushed.
As for the autoformatting, you need to run it from the root of the repo. Should possibly add some code to cd to the git root.

wheybags:
Thanks for the PR!

Exairnous:
Your welcome :smiley:
I Didn't know about the disable warning headers.  Neat!  I'll remember them for next time.
About the autoformatting, for some reason I thought it was supposed to affect the pull request on Github, not my local copy that I would then have to push to Github.  I think I've got it now :crossed_fingers:
But, I'm still a little concerned that curl can't verify the certificate.

wheybags:
Yeah, I'm not sure why that happened. It started happening for me too lately, it used to be fine.
SSL cert is still accepted by my browser, so I dunno what's going on tbh.

PR #414, opened 2019-01-09
Update player graphics when equipment changed

grantramsay:
Most the hard work was already done, just connected it up.
Can be a slight lag since the other graphics files aren't preloaded.

e.g. Mace, armour and shield:
untitled


grantramsay:
Superseded by #415, needed to expand on this small update there

PR #415, opened 2019-01-09
Play inventory sounds and update player graphics

grantramsay:
Now plays all the inventory grab/place/flip sounds.
Pulls lookup tables from Diablo.exe.

Also now updates player graphics when equipment changed.
Can be a slight lag since the other graphics files aren't preloaded.
e.g. Mace, armour and shield:
50888157-e3eb3c00-1459-11e9-8f01-a84f20015448

### DIFF
--- a/apps/freeablo/faworld/inventory.h
+++ b/apps/freeablo/faworld/inventory.h
@@ -136,15 +136,15 @@ namespace FAWorld
         BasicInventory mRightHand = BasicInventory(1, 1, true);
         BasicInventory mCursorHeld = BasicInventory(1, 1, true);
 
-        const std::vector<std::pair<EquipTargetType, BasicInventory&>> mInventoryTypes = {{EquipTargetType::inventory, mMainInventory},
-                                                                                          {EquipTargetType::belt, mBelt},
-                                                                                          {EquipTargetType::head, mHead},
-                                                                                          {EquipTargetType::body, mBody},
-                                                                                          {EquipTargetType::leftRing, mLeftRing},
-                                                                                          {EquipTargetType::rightRing, mRightRing},
-                                                                                          {EquipTargetType::leftHand, mLeftHand},
-                                                                                          {EquipTargetType::rightHand, mRightHand},
-                                                                                          {EquipTargetType::amulet, mAmulet},
-                                                                                          {EquipTargetType::cursor, mCursorHeld}};
+        std::map<EquipTargetType, BasicInventory&> mInventoryTypes = {{EquipTargetType::inventory, mMainInventory},
+                                                                      {EquipTargetType::belt, mBelt},
+                                                                      {EquipTargetType::head, mHead},
+                                                                      {EquipTargetType::body, mBody},
+                                                                      {EquipTargetType::leftRing, mLeftRing},
+                                                                      {EquipTargetType::rightRing, mRightRing},
+                                                                      {EquipTargetType::leftHand, mLeftHand},
+                                                                      {EquipTargetType::rightHand, mRightHand},
+                                                                      {EquipTargetType::amulet, mAmulet},
+                                                                      {EquipTargetType::cursor, mCursorHeld}};
     };
 }
--- a/apps/freeablo/faworld/inventory.h
+++ b/apps/freeablo/faworld/inventory.h
@@ -69,6 +69,8 @@ namespace FAWorld
         const_iterator begin() const { return mInventoryBox.begin(); }
         const_iterator end() const { return mInventoryBox.end(); }
 
+        boost::signals2::signal<void (Item const& removed, Item const& added)> inventoryChanged;
+
     private:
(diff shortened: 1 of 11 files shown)


apps/freeablo/faworld/inventory.h
```diff
@@ -115,7 +118,8 @@ namespace FAWorld
 
     public:
         // This is not serialised - it should be reconnected by other means
-        boost::signals2::signal<void()> equipChanged;
+        boost::signals2::signal<void(BasicInventory const& inventory, Item const& removed, Item const& added)> inventoryChanged;
+        boost::signals2::signal<void(Item const& removed, Item const& added)> cursorItemChanged;
```
wheybags: These should use the member prefix (mCursorItemChanged)

apps/freeablo/faworld/inventory.h
```diff
@@ -69,6 +69,8 @@ namespace FAWorld
         const_iterator begin() const { return mInventoryBox.begin(); }
         const_iterator end() const { return mInventoryBox.end(); }
 
+        boost::signals2::signal<void(Item const& removed, Item const& added)> inventoryChanged;
+
```
wheybags: Should use member prefix here

components/diabloexe/diabloexe.cpp
```diff
@@ -157,6 +158,32 @@ namespace DiabloExe
         }
     }
 
+    void DiabloExe::loadSoundFilenames(FAIO::FAFileObject& exe, size_t codeOffset)
+    {
+        uint64_t offset = mSettings->get<uint64_t>("Sounds", "filenameTable");
+        soundFilename.resize(858);
```
wheybags: What is 858? Should this magic number be in the ini file?

components/diabloexe/diabloexe.h
```diff
@@ -89,6 +90,9 @@ namespace DiabloExe
         std::vector<Affix> mAffixes;
         std::vector<std::vector<int32_t>> mTownerAnimation;
         std::vector<std::string> itemDropGraphicsFilename;
+        std::vector<std::string> soundFilename;
+        std::vector<uint32_t> itemGraphicsIdToDropSfxId;
+        std::vector<uint32_t> itemGraphicsIdToInvPlaceSfxId;
```
wheybags: Should use member prefix here

apps/freeablo/faworld/player.cpp
```diff
@@ -54,7 +54,26 @@ namespace FAWorld
...
+            (void)removed;
+
+            // Update player graphics.
+            updateSprites();
+
+            if (!added.isEmpty())
+            {
+                // Play inventory place sound.
+                std::string soundPath = added.getInvPlaceSoundPath();
+                Engine::ThreadManager::get()->playSound(soundPath);
```
wheybags: Two checks need to be added here:
1. We shouldn't play these when it's the initial items being added during character creation (maybe add a "silent" flag to inv transfers somehow).
2. We should only play these when the player is the local player (== FAWorld::World::getCurrentPlayer)

apps/freeablo/faworld/player.cpp
```diff
@@ -54,7 +54,26 @@ namespace FAWorld
...
+        mInventory.inventoryChanged.connect([this](BasicInventory const& inventory, Item const& removed, Item const& added) {
+            (void)inventory;
+            (void)removed;
+
+            // Update player graphics.
+            updateSprites();
+
+            if (!added.isEmpty())
+            {
+                // Play inventory place sound.
```
grantramsay: If inventories knew what type they were this would be much cleaner and wouldn't require the additional `cursorItemChanged` signal.
e.g.
```
switch (inventory.getType())
{
case InventoryType::body:
case InventoryType::leftHand:
case InventoryType::rightHand:
    // Update player graphics.
    updateSprites();
    break;
}

if (!added.isEmpty() /* && !silent && player == currentPlayer */ )
{
    // Play inventory place/grab sound.
    switch (inventory.getType())
    {
    case InventoryType::cursor:
        Engine::ThreadManager::get()->playSound("sfx/items/invgrab.wav");
        break;
    default:
        std::string soundPath = added.getInvPlaceSoundPath();
        Engine::ThreadManager::get()->playSound(soundPath);
        break;
    }
}
```
I might have a look at doing this refactor as well

apps/freeablo/faworld/player.h
```diff
@@ -55,6 +55,8 @@ namespace FAWorld
         int getDamageBonus() const { /* placeholder */ return 0; }
         ItemBonus getItemBonus() const;
 
+        bool mPlayerInitialised = false;
```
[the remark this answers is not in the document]
grantramsay: Sweet, added comment and assert, was already set to true during loading

wheybags:
Sorry, I thought I'd merged this ages ago!
Thanks again.

PR #416, opened 2019-01-10
Add player levelling and monster drops

grantramsay:
Partial of implementation of #48, makes the game a bit more playable

### DIFF
--- a/apps/freeablo/faworld/gamelevel.h
+++ b/apps/freeablo/faworld/gamelevel.h
@@ -93,9 +93,7 @@ namespace FAWorld
         bool isPassableFor(const Misc::Point& point, const Actor* actor) const;
 
         bool dropItem(std::unique_ptr<Item>&& item, const Actor& actor, const Tile& tile);
-
-        // If direction is NULL attempts destination point first.
-        bool dropItemClosestEmptyTile(Item& item, const Actor& actor, const Misc::Point& position, Misc::Direction* direction);
+        bool dropItemClosestEmptyTile(Item& item, const Actor& actor, const Misc::Point& position, Misc::Direction direction);
 
         Actor* getActorById(int32_t id);
 
--- a/apps/freeablo/faworld/gamelevel.h
+++ b/apps/freeablo/faworld/gamelevel.h
@@ -95,7 +95,7 @@ namespace FAWorld
         bool dropItem(std::unique_ptr<Item>&& item, const Actor& actor, const Tile& tile);
 
         // If direction is NULL attempts destination point first.
-        bool dropItemClosestEmptyTile(Item& item, const Actor& actor, const Misc::Point& position, Misc::Direction *direction);
+        bool dropItemClosestEmptyTile(Item& item, const Actor& actor, const Misc::Point& position, Misc::Direction* direction);
 
         Actor* getActorById(int32_t id);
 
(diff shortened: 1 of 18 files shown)


grantramsay:
Apologies for the big PR, got a bit carried away...

apps/freeablo/faworld/player.cpp
```diff
@@ -350,4 +350,56 @@ namespace FAWorld
...
+        // Increase HP/Mana.
+        switch (mPlayerClass)
+        {
+        case PlayerClass::warrior:
+            mStats.mHp.max += 2;
+            mStats.mMana.max += 1;
+            break;
+        case PlayerClass::rogue:
+            // TODO: Check this is correct, the below forum reckons +1.5 each.
+            // https://www.diabloii.net/forums/threads/what-is-the-max-level-on-d1.744752/#post-7322937
```
wheybags: The +2, +2 you have in the code is correct, according to [Jarulf's Guide v.162](http://www.lurkerlounge.com/diablo/jarulf/jarulf162.pdf), section 2.1.1

apps/freeablo/faworld/monster.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+        mBehaviour.reset(new BasicMonsterBehaviour(this));
+        mFaction = Faction::hell();
+        mName = monsterStats.monsterName;
+        mSoundPath = monsterStats.soundPath;
+        mStats.mHp.max = monsterStats.maxHp;
+        mStats.mHp.current = rng.randomInRange(monsterStats.minHp, monsterStats.maxHp);
+    }
+
+    Monster::Monster(World& world, FASaveGame::GameLoader& loader) : Actor(world, loader), mMonsterStats(loader)
```
wheybags: You need to add a way to tell that the actor being loaded from a save is a monster.
We do this by registering a type here https://github.com/wheybags/freeablo/blob/606d4b98ab5a28b4ad98796410034121346171a3/apps/freeablo/faworld/world.cpp#L107

You should be able to look at what we're doing for player and do the same. To test, just save and load in a dungeon.

apps/freeablo/faworld/playerstats.cpp
```diff
@@ -1,6 +1,7 @@
 #include "playerstats.h"
 #include "../fasavegame/gameloader.h"
 #include "diabloexe/characterstats.h"
+#include <algorithm>
```
wheybags: Was this intentional?

apps/freeablo/faworld/gamelevel.h
```diff
@@ -94,6 +94,9 @@ namespace FAWorld
 
         bool dropItem(std::unique_ptr<Item>&& item, const Actor& actor, const Tile& tile);
 
+        // If direction is NULL attempts destination point first.
+        bool dropItemClosestEmptyTile(Item& item, const Actor& actor, const Misc::Point& position, Misc::Direction *direction);
```
wheybags: I would add a "None" value to the enum, instead of using a pointer here.

apps/freeablo/faworld/playerstats.cpp
grantramsay: Yeah fixed windows build (`error C2039: 'min': is not a member of 'std'`), probably to do with the `NOMINMAX` flag

wheybags:
Thanks for the PR!

Issue #45, opened 2014-02-23
magic


wheybags:
This will probably need to be split into a bunch of dub-issues for each type of magic

PR #419, opened 2019-01-16
Fix/45/fixed point

grantramsay:
As mentioned in #45.
If someone can find some decent FixedPoint trig approximations feel free to update

### DIFF
--- a/components/misc/fixedpoint.cpp
+++ b/components/misc/fixedpoint.cpp
@@ -29,10 +29,13 @@ FixedPoint::FixedPoint(const std::string& str)
     release_assert(split.size() <= 2);
     release_assert(split.size() > 0);
 
+    int64_t sign = split[0][0] == '-' ? -1 : 1;
+
     int64_t integer = 0;
     {
         std::stringstream ss(split[0]);
         ss >> integer;
+        integer = i64abs(integer);
     }
     int64_t fractional = 0;
     int64_t fractionalDigits = 0;
@@ -43,7 +46,6 @@ FixedPoint::FixedPoint(const std::string& str)
         ss >> fractional;
     }
 
-    int64_t sign = integer >= 0 ? 1 : -1;
     integer = i64abs(integer);
 
     int64_t tmpScalePow10 = fractionalDigits;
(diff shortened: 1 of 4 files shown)


components/misc/fixedpoint.cpp
```diff
@@ -108,12 +114,16 @@ double FixedPoint::toDouble() const
 std::string FixedPoint::str() const
 {
     std::stringstream ss;
-    ss << intPart();
```
grantramsay: This didn't work for negatives if intPart is zero.
Similar for from string constructor.
`fractionPart` above dropped the sign, and `round` didn't work for negatives

tilkinsc:
By the way, because of the fact that floats have a mantissa, you can emulate perfectly a fixed point number with performance gains. You just got to directly edit the memory by force because compilers aren't very forgiving.

grantramsay:
Sounds like a neat way to do it. [Might not be that great for portability though :/](https://en.m.wikipedia.org/wiki/Endianness#Floating_point). Also sounds like gcc has native fixed point support and might be in the c standard soon too

tilkinsc:
It's portable. Floats aren't effected by endieness.

grantramsay:
From linked article above:
> the widespread IEEE 754 floating-point standard does not specify endianness
> one may in practice safely assume that the endianness is the same for floating-point numbers as for integers

grantramsay:
Superseded by #423

wheybags:
Derp, I just implemented a fixed point version of atan2 based on a LUT, without noticing you'd already added an implementation here. Mine is at https://github.com/wheybags/freeablo/tree/lut_atan2, I'll let you decide which is better. I did try implementing code from the same source you used, but I got pretty shitty accuracy from it, so unless you somehow just did it better (entirely possible, lol), it might be better to go with a LUT.

grantramsay:
Oh hahaha.
Under the section "QUICK NOTE FOR BETTER ACCURACY" they give the formula for increasing from 1st to 3rd order approximation which gives a pretty good fit.
Umm what I've got in there works pretty well with a bunch less LOC, only issue may be speed, those 128 bit (multiply/divide) operations are probably quite slow. Maybe we run it like this and change to a LUT later if we find it's too slow?

components/misc/fixedpoint.cpp
```diff
@@ -293,6 +292,14 @@ FixedPoint FixedPoint::sin(FixedPoint rad)
 }
 
 FixedPoint FixedPoint::cos(FixedPoint rad) { return sin(rad + PI / 2); }
-FixedPoint FixedPoint::atan2_degrees(FixedPoint y, FixedPoint x) { return atan2(x, y) * 180 / PI; }
-FixedPoint FixedPoint::sin_degrees(FixedPoint deg) { return sin(deg * PI / 180); }
+FixedPoint FixedPoint::atan2_degrees(FixedPoint y, FixedPoint x)
+{
+    static const FixedPoint radToDeg = FixedPoint(180) / PI;
+    return atan2(x, y) * radToDeg;
```
grantramsay: Interesting profiling this, the `* radToDeg` takes longer then the `atan2(x, y)`...
i.e. `atan2_degrees` takes more than twice the time of `atan2`
Screen Shot 2019-03-13 at 5 59 31 PM
`¯\_(ツ)_/¯`

wheybags:
Aaah, ok. I just gave up when I saw I was getting awful results, I didn't bother with the "slightly better" version.
Thanks for all the work! The other misc improvements to FixedPoints are really great to have too.

PR #420, opened 2019-01-17
Windows fixes

wheybags:
closes #231


resources/settings-default.ini
```diff
@@ -4,7 +4,7 @@
 [Display]
 resolutionWidth = 1280
 resolutionHeight = 960
-fullscreen=true
+fullscreen=false
```
grantramsay: Fullsceen is also broken on my windows VM, but only because the window is never **_resized_** when entering full screen mode so the window size isn't updated. Adding the last line below (render/sdl2backend.cpp) to extract the window size after creation fixes it for me:

        screen = SDL_CreateWindow(title.c_str(), 20, 20, WIDTH, HEIGHT, flags);
        if (screen == NULL)
            printf("Could not create window: %s\n", SDL_GetError());

        SDL_GetWindowSize(screen, &WIDTH, &HEIGHT);

PR #424, opened 2019-01-27
Fix window scaling issues

grantramsay:
Fullscreen freeablo worked fine, but launcher and windowed freeablo were way off.
Issue seems to be that [`SDL_GL_GetDrawableSize`](https://wiki.libsdl.org/SDL_GL_GetDrawableSize) always returns the high-DPI count. Either using `SDL_GetWindowSize` instead or enabling high-DPI (`SDL_WINDOW_ALLOW_HIGHDPI`) fixes the issue.
e.g. before:
screen shot 2019-01-27 at 5 09 31 pm screen shot 2019-01-27 at 5 09 40 pm
after:
screen shot 2019-01-27 at 5 09 56 pm screen shot 2019-01-27 at 5 10 03 pm

### DIFF
--- a/components/render/nuklear_sdl_gl3.cpp
+++ b/components/render/nuklear_sdl_gl3.cpp
@@ -188,7 +188,14 @@ void nk_sdl_render_dump(Render::SpriteCacheBase* cache, NuklearFrameDump& dump, 
         {-1.0f, 1.0f, 0.0f, 1.0f},
     };
     SDL_GetWindowSize(win, &width, &height);
-    SDL_GL_GetDrawableSize(win, &display_width, &display_height);
+
+    // Note: if SDL_GL_GetDrawableSize() is used SDL_WINDOW_ALLOW_HIGHDPI option must
+    // also be enabled. However this seems to cause an issue where black lines appear
+    // between each tile on high-DPI displays so is currently disabled.
+    // SDL_GL_GetDrawableSize(win, &display_width, &display_height);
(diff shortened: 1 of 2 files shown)


grantramsay:
Changed fix to not use high-DPI, as I was seeing black lines between each tile when enabled.
I might have a bit more of a look into why it doesn't work, maybe just rounding or off-by-one..
e.g. enabled:
screen shot 2019-01-27 at 5 57 32 pm
disabled:
screen shot 2019-01-27 at 6 00 12 pm

wheybags:
High DPI scaling is a bit of an issue in general, especially making it work properly across multiple platforms.
We have a high DPI mac in work, and I have 4k windows/linux at home. Before 0.4 is released, I'll have a look and try get something together that works consistently on all three (probably from this PR, or a slightly modified version of it). For now (0.4) I'm happy to just have non-corrupted rendering, but eventually I'd like to add decent scaling, where we can scale the gui and game independently.

grantramsay:
Ok sounds good. I also applied the fix I mentioned [here](https://github.com/wheybags/freeablo/pull/420#discussion_r251223077) for Windows in full screen mode.
Feel free to update/edit this PR when doing your portability testing

MaxDesiatov:
This PR fixes macOS window scaling for me, without I see this:

wheybags:
Ok, finally tested on everything, and seems to work fine.
Thanks!

PR #425, opened 2019-01-27
Allow using newer versions of clang-format

grantramsay:
Newer clang-format was touching namespace comments on every file..
Tested on 8.0.0, CI build currently uses 5.0.0
Other file changes are whitespace only from running it, and hopefully back compatible with 5.0.0

Edit: Ugh 8.0.0 doesn't seem to be back compatible, unless there's some additional formatting options for those changes. Oh well it's still better than it was

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -58,6 +58,9 @@ AlignAfterOpenBracket: Align
 DerivePointerAlignment: false
 PointerAlignment: Left
 
+# misc comment settings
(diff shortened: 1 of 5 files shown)


wheybags:
If you want to add a PPA or something similar to travis.yml so we get the newer version there, then that would be cool.
Otherwise, I think we should just stick to the version we currently have available, so I'll close this for now.

PR #426, opened 2019-02-08
Fix occasional crash from placing wall stairs in an invalid location

grantramsay:
`Dun::pointIsValid` was not being checked when generating level stairs
Added a helper placement constraint checking function

### DIFF
--- a/apps/freeablo/falevelgen/levelgen.cpp
+++ b/apps/freeablo/falevelgen/levelgen.cpp
@@ -79,6 +79,17 @@ namespace FALevelGen
         blank = 115
     };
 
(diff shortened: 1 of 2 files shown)


grantramsay:
FYI: Also noticed a different random crash when changing level:
`ASSERTION FAILED: (!blocking) in ../apps/freeablo/faworld/gamelevel.cpp:118`
I wasn't debugging at the time, maybe a monster walked onto the stairs?

wheybags:
Great, thanks!
As for the other crash, well, hopefully someone will run into it with a debugger attached and find a fix  `¯\_(ツ)_/¯`

PR #429, opened 2019-03-02
Atlas texture and batched rendering

grantramsay:
Spent ages trying to get this working with a `GL_TEXTURE_2D_ARRAY` but couldn't quite get it to work, parts of the Nuklear GUI would be missing if the array texture was allocated with > 1 depth (layers).
Currently it uses an array of textures (up to 8) to store all the images, runs at ~130 FPS on my machine (`GL_TEXTURE_2D_ARRAY` runs at ~180).
Needs a test on a few different PCs


components/render/atlastexture.cpp
```diff
@@ -0,0 +1,154 @@
...
+        //                                                              A A A A A A A A A A A A A A A A
+        static const int32_t PADDING = 1;   // Must be >= 0
+        static const int32_t ALIGNMENT = 1; // Must be >= 1
+        int32_t paddedWidth = (width + PADDING + ALIGNMENT - 1) / ALIGNMENT * ALIGNMENT;
+        int32_t paddedHeight = (height + PADDING + ALIGNMENT - 1) / ALIGNMENT * ALIGNMENT;
+
+        // This assert may cause trouble as one of the Diablo 1 images is 10752 pixels high.
+        // However ~46% (https://feedback.wildfiregames.com/report/opengl/feature/GL_MAX_TEXTURE_SIZE)
+        // of graphics cards are limited to 8192, not quite sure how it ever worked..
+        release_assert(paddedWidth <= mTextureWidth && paddedHeight <= mTextureHeight); // Texture size too small...
```
grantramsay: This comment mentions my main portability concern

resources/shaders/basic.frag
```diff
@@ -1,27 +1,28 @@
...
+
 out vec4 frag_colour;
-uniform sampler2D tex;
-uniform float h_color_r;
-uniform float h_color_g;
-uniform float h_color_b;
-uniform float h_color_a;
-uniform float imgW;
-uniform float imgH;
+uniform sampler2D tex[8]; // NOTE: The size of this array has a large impact on performance...
```
grantramsay: This array size has the main impact on performance.
With a `GL_MAX_TEXTURE_SIZE` of 16384 this only needs to be 2, 8192 is still common though so went with 8.
Using a `GL_TEXTURE_2D_ARRAY` avoids the need for this array (`uniform sampler2DArray tex;`)

components/render/sdl2backend.cpp
```diff
@@ -87,10 +151,18 @@ namespace Render
...
         SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
 
         glContext = SDL_GL_CreateContext(screen);
+
+        // Check opengl version is at least 3.3.
+        const GLubyte* glVersion(glGetString(GL_VERSION));
+        int major = glVersion[0] - '0';
+        int minor = glVersion[2] - '0';
+        if (major < 3 || (major == 3 && minor < 3))
+            std::cerr << "ERROR: Minimum OpenGL version is 3.3. Your current version is " << major << "." << minor << std::endl;
```
wheybags: I woudl abort() here. Maybe this whole print could be a message passed to release_assert.

components/render/sdl_gl_funcs.h
```diff
@@ -3,6 +3,15 @@
 #ifndef SDL_GL_FUNCS
 #define SDL_GL_FUNCS
 
+#define GL_CHECK_ERROR()                                                                                                                                       \
+    {                                                                                                                                                          \
+        GLenum err;                                                                                                                                            \
+        while ((err = glGetError()) != GL_NO_ERROR)                                                                                                            \
+        {                                                                                                                                                      \
+            fprintf(stderr, "glError %s:%d, 0x%04X\n", __FILE__, __LINE__, err);                                                                               \
```
wheybags: Maybe insert a debug break here, like we have in debug/release_assert?

apps/freeablo/farender/spritecache.cpp
```diff
@@ -257,6 +257,9 @@ namespace FARender
 
     void SpriteCache::evict()
     {
+        if (!Render::SpriteGroup::canDelete())
+            return;
```
wheybags: I think, we can get rid of the whole concept of a Sprite Cache. 
I woule be happy with just preloading all the textures we're going to use into the atlas at startup, and if we have problems with this later, then we can figure it out later.

components/render/atlastexture.cpp
```diff
@@ -0,0 +1,153 @@
...
+        glBindFramebuffer(GL_DRAW_FRAMEBUFFER, fbo);
+
+        for (int i = 0; i < mTextureLayers; i++)
+        {
+            bind(i);
+
+            // Allocate memory for texture array (by passing NULL).
+            // NOTE: For GL_COMPRESSED_RGBA image dimensions need to be padded to a
+            // multiple/alignment of 4: https://forums.khronos.org/showthread.php/77554
+            // Also may need to clear texture if there are
```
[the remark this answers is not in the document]
grantramsay: Ahh yep, started writing a "TODO" then just did it..

apps/freeablo/farender/spritecache.cpp
grantramsay: Ahh yep sweet, will give this a go.
That method name is also terrible, was meant to be `canDeleteSprites`

components/render/atlastexture.cpp
grantramsay: "ui_art/font42g.pcx"
Exported it to jpeg:
font42g

apps/freeablo/farender/spritecache.cpp
wheybags: We could dump a list, by running the game and tracing each load. Or, honestly, we could just preload every CEL/CL2 file in DIABDAT.MPQ.

components/render/atlastexture.cpp
wheybags: Looks like it does work, using Render::loadResizedSprite. So, problem solved :D

apps/freeablo/farender/spritecache.cpp
grantramsay: @wheybags I had a go at preloading all the sprites, it seems there are simply too many.
My debug outputs: 1402 Cel files, 110247 individual sprites.
The 2D bin packer mustn't have a very good Big O and gets slower and slower, but even disabling that just the Cel decoding takes > 60s.

I don't see dumping a list of _used_ Cel files as being very maintainable.
Might have to stick with loading files runtime?

Also I had to disable decoding these 3 Cel files as they cause a crash when decoding (they also crash if clicked in the celview executable):
```
"monsters/fireman/firema.cl2"
"monsters/unrav/unravw.cel"
"monsters/darkmage/dmagew.cl2"
```

AJenbo:
The 3 mentioned cel files are corrupt so won't decode under normal conditions

components/render/atlastexture.cpp
grantramsay: Just requires:
- updating `Renderer::loadFonts()` to pass the additional info (`resize=640x672&tileSize=40x42`)
- `SpriteCache::get(const std::string& path)` also needs to handle `pair[0] == "resize"` etc like `SpriteCache::get(uint32_t index)` (can probably make the functionality common to both).
- `PcxFontInfo` needs to be given the extra information ("tile" size, font height (Note: font height 24 actually 26, 30 actually 31), texture width/height).
- `PcxFontInfo::queryGlyph` needs to be updated to return the correct uv coorinates. Note: resizing the font image also transposes the sub images in this case i.e.
    ```
    1         resize ->     1    2    3    4
    2                       5    6    7    8
    3                       9   ....
    4
    ...

grantramsay:
Ahh thanks @AJenbo that makes sense

components/render/atlastexture.cpp
```diff
@@ -0,0 +1,148 @@
...
+        GLint maxTextures = 8; // Hardcoded in fragment shader, GL 3.x minimum is 16.
+        glGetIntegerv(GL_MAX_TEXTURE_SIZE, &maxTextureSize);
+
+        mTextureWidth = maxTextureSize;
+        mTextureHeight = maxTextureSize;
+
+        // Limit number of textures to a reasonable level (measured from testing).
+        // Note: Increasing this has a severe impact on performance.
+        GLint estimatedRequiredTextures = (1uLL << 29) / (mTextureWidth * mTextureHeight);
+        mTextureLayers = std::min(estimatedRequiredTextures, maxTextures);
```
vieiraa: I'm getting maxTextureSize = 32768 on my GTX 1050. This makes mTextureWidth * mTextureHeight be equal to 1073741824, which makes the division in estimatedRequiredTextures be equal to 0, making mTextureLayers be equal to 0. This leads to a assertion failure (release_assert(layer < mTextureLayers)).

vieiraa:
```suggestion
        mTextureLayers = estimatedRequiredTextures > 0 ? std::min(estimatedRequiredTextures, maxTextures) : maxTextures;
```

vieiraa: Doing this leads to a glError (freeablo/components/render/sdl2backend.cpp:694, 0x0506)

grantramsay: Try `1` instead of `maxTextures`, e.g:

    mTextureLayers = estimatedRequiredTextures > 0 ? std::min(estimatedRequiredTextures, maxTextures) : 1;

It probably should just round up, e.g:

    // In some helper header
    #define DIVIDE_ROUND_UP(numerator, demoninator) \
            ( ((numerator) + ((demoninator) - 1u)) / (demoninator) )
    
    GLint estimatedRequiredTextures = DIVIDE_ROUND_UP(1uLL << 29, mTextureWidth * mTextureHeight);
C++ probably has build in functionality for this

vieiraa: Now it is allocating so much memory it freezes my system (I have 8gb of RAM and a 19gb swap partition).

grantramsay: It might be allocating too much vram from your graphics card..
Can you try this instead:

glGetIntegerv(GL_MAX_TEXTURE_SIZE, &maxTextureSize);
maxTextureSize = std::min(maxTextureSize, 16384);
 mTextureWidth = maxTextureSize;

vieiraa: Now it worked. Getting 370-400 fps. Wow.

wheybags:
I get the error
> 0:14(19): error: sampler arrays indexed with non-constant expressions are forbidden in GLSL 1.30 and later

If you have time, I think the proper answer is either to move to texture arrays, or break the batch when the atlas texture changes. The second option is more scalable IMO, but it would require some intelligent allocation of atlases (at the very least, the ground textures should all be in one atlas).

wheybags:
I will get around to renderer work eventually myself, if I do before this is merged, I'll just start working on this branch myself. Either way, what's here already is a good chunk of what needs to be done, so thanks a lot!

apps/freeablo/farender/spritecache.cpp
wheybags: What I would do here is keep some list of sprites, and on first run, decode them all into atlases, and save the atlases as pngs in soem temp folder. Then we can just laod the single pngs on subsequent runs. OFC we would need to invalidate that if the sprite list changes.

grantramsay:
Yeah there's some good stuff in here but it's not stable enough to merge. Storing all the textures is a bit heavy on gpu ram. Splitting textures into drawing layer categories (if possible) and only caching the current level tiles would probably be better. All good, hopefully what's here helps you out

grantramsay:
I had a thought that clearing the atlas texture when changing levels would be a simple way to reduce the large RAM requirements of storing all the sprites, decided to give it a go..
Uses a 2D array texture 8192*8192 with 2 layers.
It's just a thought, let me know what you think. And feel free to discard if you carry this work on in a different direction

wheybags:
That's a direction we could go, but tbh I think we should be ok to just store the whoel game's assets in VRAM. factorio does this (normally, it can swap out too if you're low on VRAM, but that's a relatively recent addition) and we should have waaaaaaaay less image data to store.

wheybags:
Im going to start wokring on a bit of an overhaul to the rendering code, so I think I will just merge this as-is for now, and use it as a base (also avoids my pr getting absolutely massive if I just branch off this and merge it all together :v)

PR #439, opened 2019-06-10
Fix mistake in heuristic

akuskis:
Hi @wheybags , please review.

Fixed bug in heuristic.

Before:
```
. . ✔ . . 
. ✔ . . . 
✔ . . . . 
✔ . . . . 
. ✔ . . . 
. . ✔ . .
```

After:
```
. . ✔ . . 
. . ✔ . . 
. . ✔ . . 
. . ✔ . . 
. . ✔ . . 
. . ✔ . .
```

### DIFF
--- a/apps/freeablo/faworld/findpath.cpp
+++ b/apps/freeablo/faworld/findpath.cpp
@@ -66,7 +66,7 @@ namespace FAWorld
         auto dx = abs(b.x - a.x);
         auto dy = abs(b.y - a.y);
 
-        auto straight = static_cast<size_t>(dx - dy);
+        auto straight = std::abs(dx - dy);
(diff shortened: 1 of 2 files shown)


akuskis:
Fix for #440

wheybags:
Thanks for the fix!

AJenbo:
Just some trivia, in the original game, if A and B were not aligned on either x or y the path would first move diagonally until they align. This appears to be down to the order the A* implementation looks at different directions on the grid:
```
-1,-1 = NW
-1, 1 = SW
 1,-1 = SE
 1, 1 = NW
-1, 0 = W
 0,-1 = N
 1, 0 = E
 0, 1 = S
```

PR #444, opened 2019-10-14
Start on proper combat formulas

wheybags:


apps/freeablo/faworld/gamelevel.cpp
```diff
@@ -86,11 +86,11 @@ namespace FAWorld
 
     void GameLevel::update(bool noclip)
     {
-        for (size_t i = 0; i < mActors.size(); i++)
-        {
-            Actor* actor = mActors[i];
+        for (auto& actor : mActors)
+            actor->recalculateStats();
+
+        for (auto& actor : mActors)
```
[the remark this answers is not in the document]
wheybags: No, because a damage calculation later on might depend on the stats of another actor whose stats haven't been refreshed yet. Eg: I damage you, I've been refreshed, you haven't, so maybe your armor value is incorrect because you just changed your gear in the last tick. It would be only one frame off, but might as well be correct and avoid surprises.

apps/freeablo/faworld/player.cpp
```diff
@@ -100,34 +100,110 @@ namespace FAWorld
...
+        stats.baseStats.strength = charStats.strength + itemStats.baseStats.strength;
+        stats.baseStats.magic = charStats.magic + itemStats.baseStats.magic;
+        stats.baseStats.dexterity = charStats.dexterity + itemStats.baseStats.dexterity;
+        stats.baseStats.vitality = charStats.vitality + itemStats.baseStats.vitality;
+
+        stats.toHitMelee.bonus = 0;
+        stats.toHitRanged.bonus = 0;
+        stats.toHitMagic.bonus = 0;
+
+        // TODO: make sure all the following calculations should be rounded
```
wheybags: @AJenbo can you tell me how this works in vanilla? Are the values for damage, hit chance etc rounded before use, or are damage calculations done with floating point?

vieiraa: This guide has the game's combat formulas:
http://www.lurkerlounge.com/diablo/jarulf/jarulf162.pdf

PR #445, opened 2019-11-23
Allow compilation with VS2017

vieiraa:
I couldn't make it work with VS2019 though.
Fix warning in diabloexe.cpp.


wheybags:
Thanks!

PR #447, opened 2019-11-24
Add buy/sell confirm dialog

### LINKED ISSUE #395: Buy/sell confirm dialog
Buying and selling should open a confirmation dialog

vieiraa:
Fixes #395 

image

image

### DIFF
--- a/apps/freeablo/fagui/shopdialogs.h
+++ b/apps/freeablo/fagui/shopdialogs.h
@@ -31,7 +31,7 @@ namespace FAGui
             sell
         };
 
-        ConfirmTransactionPopup(GuiManager& guiManager, const std::string& intro, const std::vector<std::string>&& desc, Transaction t);
+        ConfirmTransactionPopup(GuiManager& guiManager, const std::string& intro, const std::vector<std::string>& desc, Transaction t);
         void addAction(std::function<CharacterDialoguePopup::UpdateResult()> action);
 
     protected:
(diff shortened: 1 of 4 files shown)


apps/freeablo/fagui/shopdialogs.h
```diff
@@ -47,10 +69,10 @@ namespace FAGui
...
     private:
-        void buyItem(size_t index);
+        void buyItem(size_t index, ConfirmTransactionPopup* confirmPopup);
 
     private:
         std::vector<FAWorld::StoreItem>& mItems;
         const FAWorld::Actor& mShopkeeper;
     };
-}
+} // namespace FAGui
```
wheybags: Nitpick, but I'm not a fan of these end namespace comments. I presume they were added by your IDE?

AJenbo: It's defined in your .clang-format, see FixNamespaceComments in https://clang.llvm.org/docs/ClangFormatStyleOptions.html
Apply preference here:
https://github.com/wheybags/freeablo/blob/master/.clang-format

wheybags: From what I can tell, we want that option off, and it should be off by default since we didn't enable it? Unless I'm missing something?

AJenbo: Afaik it's on by default

wheybags: Well, either way, I just set it to false explicitly now.

vieiraa: It was added by Visual Studio, sorry about that.

vieiraa: It was added by Visual Studio, sorry about that.

AJenbo: Visual Studio **Code** uses clang-format afaik.

PR #448, opened 2019-12-13
Add MD5 validation

vieiraa:
Fixes #427

PR #450, opened 2020-01-23
Remove Boost MD5 library usage.

vieiraa:
Follow up on #448.

### DIFF
--- a/apps/freeablo/fa_main.cpp
+++ b/apps/freeablo/fa_main.cpp
@@ -116,7 +116,12 @@ int fa_main(int argc, char** argv)
     // Check if we've been configured with data files, and if we haven't, run the launcher to prompt configuration
     if (!(settings.loadUserSettings() && dataFilesSetUp(settings)))
     {
-        system((boost::filesystem::system_complete(argv[0]).parent_path() / "launcher").string().c_str());
+#ifdef _WIN32
+        std::string path = "\"" + (boost::filesystem::system_complete(argv[0]).parent_path() / "launcher").string() + "\"";
+#else
+        std::string path = (boost::filesystem::system_complete(argv[0]).parent_path() / "launcher").string();
+#endif
+        system(path.c_str());
         return EXIT_SUCCESS;
     }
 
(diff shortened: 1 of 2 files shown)


apps/freeablo/fa_main.cpp
```diff
@@ -103,7 +100,7 @@ bool dataFilesSetUp(const Settings::Settings& settings)
 
     for (const auto& hash : hashes)
     {
-        mpqValid = md5Equal(hash, mpqDigest);
+        mpqValid = md5Equal(hash, digest);
```
grantramsay: Looks like it only checks if the last (eu version) hash is correct..

apps/launcher/main.cpp
```diff
@@ -200,7 +200,14 @@ int main(int, char** argv)
     }
 
     if (runFreeablo)
-        system((boost::filesystem::system_complete(argv[0]).parent_path() / "freeablo").string().c_str());
+    {
+#ifdef _WIN32
+        std::string path = "\"" + (boost::filesystem::system_complete(argv[0]).parent_path() / "freeablo").string() + "\"";
```
grantramsay: Is this for spaces in directory path?
If so:
    It might have same issue on 64 bit windows and posix..
    `boost::filesystem` probably has something portable for escaping spaces

vieiraa: The closest thing I could find would be to use a regex to escape the spaces. Is this ok?

vieiraa: And I don't believe POSIX has this issue. I'll check this later.

apps/freeablo/fa_main.cpp
vieiraa: What do you mean? It's working here.

grantramsay: Looks like it needs `if (mpqValid) break;` in that loop

apps/launcher/main.cpp
grantramsay: That's odd, posix shell should require spaces escaped too..
Regex is a bit overkill.
Maybe just escape path for any OS. otherwise nevermind

vieiraa: Actually, you're right. Done.

wheybags:
Thanks for the PR!

apps/launcher/main.cpp
wheybags: Did you test this on windows? Won't the \\ show up as a directory separator there?

vieiraa: Ugh. I'm gonna open another pr for that.

PR #452, opened 2020-01-24
Properly escape paths on Windows

vieiraa:


wheybags:
Works just fine for me on windows. Thanks!

PR #454, opened 2020-01-25
Create talk dialog

vieiraa:
Added addresses for gossip and quest dialogs.

TODO:

- [x] Implement text scroll dialog.
- [x] Add NPC voices.


vieiraa:
So far it looks like this:

image

image

I'm gonna need help with the autoscroll. I have no idea how to do it.

wheybags:
Nice!
I can't test the audio part ATM, as I'm on a train with no heaphones, but it seems like it shoudl work.
Thanks for the PR!

PR #458, opened 2020-02-06
No boost

wheybags:


grantramsay:
Nice!
Just noticed a couple occurrences left after clearing out hunter (`rm -rf ~/.hunter`) and rebuild.
https://github.com/wheybags/freeablo/search?q=boost+-path%3Aextern

wheybags:
ah damn, I was searching for <boost/ to find all occurrences, didn't pick up on any includes using quotes :p
Will fix now.

PR #462, opened 2020-02-14
Fix/select hero menu crash

grantramsay:
Fix crash when enter hit twice on SelectHeroMenuScreen.
Also fix some Clang-Tidy (static analysis) warnings in EngineMain


wheybags:
Thanks!

Issue #392, opened 2018-11-10
Click and drag off gui should not pass through to game


wheybags:
When we click on the gui, we suppress the event from going to the game.
When we click down on the gui, then drag off the gui, these events should be supressed too.

wheybags:
closed by #464

PR #464, opened 2020-02-15
Fix/392/drag off gui

grantramsay:
Fix for #392

### DIFF
--- a/apps/freeablo/engine/engineinputmanager.cpp
+++ b/apps/freeablo/engine/engineinputmanager.cpp
@@ -163,10 +163,10 @@ namespace Engine
     {
         NuklearMisc::handleNuklearMouseMoveEvent(mNkCtx, x, y, xrel, yrel);
 
+        mMousePosition = Misc::Point{x, y};
+
         if (!nk_item_is_any_active(mNkCtx))
             notifyMouseObservers(MouseInputAction::MOUSE_MOVE, mMousePosition, mMouseDown, mKbMods);
-
-        mMousePosition = Misc::Point{x, y};
     }
 
     std::string EngineInputManager::keyboardActionToString(KeyboardInputAction action) const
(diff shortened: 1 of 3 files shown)


apps/freeablo/engine/engineinputmanager.cpp
```diff
@@ -123,31 +123,35 @@ namespace Engine
...
-        {
-            mMousePosition = Misc::Point{x, y};
-            mMouseDown = true;
-            mClick = true;
-        }
-        if (key == Input::Key::KEY_RIGHT_MOUSE)
+        mMousePosition = Misc::Point{x, y};
+
+        // Only pass relevant mouse clicks to game engine.
+        if (!mPaused && !nk_item_is_any_active(mNkCtx) && !mGuiManager->isModalDlgShown())
```
grantramsay: `mMouseDown`/`mRightMouseDown` etc are now only relevant for the game engine.
The other option was to add additional state e.g. `mMouseDownIgnore`/`mRightMouseDownIgnore` which seemed worse.
Seems to work well, let me know if I've missed/broken anything

apps/freeablo/faworld/playerbehaviour.cpp
```diff
@@ -104,7 +107,10 @@ namespace FAWorld
             case PlayerInput::Type::CastSpell:
             {
                 auto clickedPoint = Misc::Point(input.mData.dataCastSpell.x, input.mData.dataCastSpell.y);
-                mPlayer->castActiveSpell(clickedPoint);
+                if (!cursorItem.isEmpty())
```
wheybags: This doesn't make sense to me, there's a separate PlayerInput for dealing with dropping items.

grantramsay: Ahh I see, `LocalInputHandler` should do this check. Yep that would make this much clearer.

Also since adding this "cast spell input" for right click, I've found out that it's incorrect. Right click doesn't mean cast spell, it means "do whatever action you've configured it to do" (via the bottom ctrl panel). So there's a bit of refactoring needed here as well

PR #465, opened 2020-02-16
Feat/45/spells

grantramsay:
More work towards #45, still lots to do.
Added a spells layer to support existing missiles.
Implemented the spellbook panel to select active spell.


wheybags:
I couldn't comment yesterday when I merged, as the github webui was having issues, but thanks a lot, this is great!

apps/freeablo/faworld/player.h
```diff
@@ -44,15 +44,15 @@ namespace FAWorld
 
         bool castSpell(SpellId spell, Misc::Point targetPoint) override;
         void setActiveSpellNumber(int32_t spellNumber);
+        SpellId getActiveSpell() const { return mActiveSpell; }
+        void setActiveSpell(SpellId spell) { mActiveSpell = spell; }
```
wheybags: When there is both a getter and setter with no special logic, I think it makes more sense to just make the variable public.

PR #467, opened 2020-02-21
Remove unused spell hotkeys

grantramsay:
These are unused from a previous commit and need to be handled in a different way/place in final implementation


wheybags:
Cool, thanks!

Issue #469, opened 2020-02-21
Compilation issue (audio/audio.h)


akuskis:
![Selection_004](https://user-images.githubusercontent.com/3109998/75065716-64ec8200-54fa-11ea-840c-07cdc8370079.png)

File /usr/include/audio/audiolib.h is looking for "audio/audio.h" and instead of using this header from "/usr/include" it took this header from "freeablo/compoments"

PR #470, opened 2020-02-21
Reduce components visibility #469

akuskis:
This one should reduce visibility of freeablo/components, to fix issue #469


wheybags:
Would it be a better solution to just rename our header to fa_audio.h or something like that?

akuskis:
If you are fine with that, then I can rework it.
Why I didn't rename it and only reduced scope, because there can be other situations like this.

wheybags:
Well, even with the reduced scope, it can still happen that there is a collision in a module that actually needs both headers. We should be making sure our header names are un-generic enough that they're unlikely to cause collisions.

akuskis:
Updated

wheybags:
Cool, thanks!

PR #471, opened 2020-02-24
Show fps in the title bar

vieiraa:
I want to redirect stdout to the console.


wheybags:
I would like to add something better in-game (like a debug panel, I even started work on it once), but this is better than what we have for now.
Thanks!

PR #474, opened 2020-02-29
Add animations for casting spells and fire arrows for ranged weapons

### LINKED ISSUE #46: bows
firing animation and show arrows

grantramsay:
Added `RangedAttackState` and `SpellAttackState` classes (similar to MeleeAttackState) and a common ABC. Added other minor bits to enable spell casting animations and firing arrows.
Renamed "AttackDirection" input to "ForceAttack" and changed its data to a target point instead of a direction (as ranged/bow attacks require this information).
This probably almost closes #46 (Bows) too

### DIFF
--- a/apps/freeablo/faworld/actor/attackstate.cpp
+++ b/apps/freeablo/faworld/actor/attackstate.cpp
@@ -1,47 +1,107 @@
 #include "attackstate.h"
 #include "../../fasavegame/gameloader.h"
 #include "../actor.h"
+#include "../spells.h"
 
 namespace FAWorld
 {
-
     namespace ActorState
     {
-        const std::string MeleeAttackState::typeId = "actorstate-attack-state";
+        const std::string BaseAttackState::typeId = "actorstate-base-attack-state";
 
-        void MeleeAttackState::save(FASaveGame::GameSaver& saver) const { mDirection.save(saver); }
+        void BaseAttackState::save(FASaveGame::GameSaver& saver) const
+        {
+            mTargetPoint.save(saver);
+            saver.save(mAttackDone);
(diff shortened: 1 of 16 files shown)


apps/freeablo/faworld/actor/attackstate.cpp
```diff
@@ -1,47 +1,107 @@
...
 #include "../actor.h"
+#include "../spells.h"
 
 namespace FAWorld
 {
-
     namespace ActorState
     {
-        const std::string MeleeAttackState::typeId = "actorstate-attack-state";
+        const std::string BaseAttackState::typeId = "actorstate-base-attack-state";
```
wheybags: BaseAttackState is already an abstract class, so it doesn't need to have an id. It's getTypeId() method can be left unimplemented.
This is only necessary so we can tell what class to use for deserialisation, and a BaseAttackState can never be saved anyway since it can't be instantiated directly (without using a subclass)

apps/freeablo/faworld/spells.h
```diff
@@ -18,6 +26,20 @@ namespace FAWorld
 
         const std::string& soundEffect() const { return mSpellData.mSoundEffect; }
 
+        SpellType getType() const
+        {
```
wheybags: I would put this enum in the DiabloExe class, and change the type of DiabloExe::SpellData::mType to SpellType instead of an untyped int.

grantramsay:
That bug was to do with dragging mouse while shift key is down, should be fixed now
Also changed shift key priority to be higher then target actor/item click

wheybags:
Nice, thanks!

PR #475, opened 2020-03-01
Feat/missile class composition

grantramsay:
- Missile attributes/characteristics are now specified in one place, and are initialised in the missile constructor (closer to proper composition).
- Missile attributes now use nested classes instead of friendship for private member access.
- Allow setting different speeds for missiles.
- Added a reference to missiles graphics on each level to avoid having to check every actor in world to find missile graphics on a level.
- Added town portal spell.
- Put missile max range within missile movement attribute

A lot of the changes are just whitespace from the "Tidy namespaces" commit


grantramsay:
Will tidy this up and fix build later

apps/freeablo/faworld/missile/missilegraphic.cpp
```diff
@@ -19,16 +24,28 @@ namespace FAWorld::Missile
         mMainGraphicPath = loader.load<std::string>();
         mSingleFrame = (mSingleFrame = loader.load<int32_t>()) == -1 ? std::nullopt : mSingleFrame;
         mAnimationPlayer = FARender::AnimationPlayer(loader);
+        auto levelIndex = loader.load<int32_t>();
+        auto world = loader.currentlyLoadingWorld;
+        loader.addFunctionToRunAtEnd([this, world, levelIndex]() {
+            mLevel = world->getLevel(levelIndex);
+            mLevel->mMissileGraphics.push_back(this);
```
wheybags: It's not currently a probem, since we only use this list for rendering,  but the order of the list will not be the same when loading a game as it was before saving. If we ever use the list for any kind of game logic, this could be a problem. Maybe add a comment there that warns not to use it for game logic. If we ever do need them, then we can make all of the inserts sort by some serialised unique id or something similar.

apps/freeablo/faworld/gamelevel.cpp
```diff
@@ -192,7 +192,7 @@ namespace FAWorld
             Misc::Point targetPoint = point + Misc::Point{xOffset, yOffset};
             if (targetPoint.x >= 0 && targetPoint.x < width() && targetPoint.y >= 0 && targetPoint.y < height())
             {
-                if (isPassable(targetPoint, nullptr))
+                if (isPassable(targetPoint, nullptr) && (additionalConstraints == nullptr || additionalConstraints(targetPoint)))
```
wheybags: Instead of adding this special callback, it would be nice to refactor the class layout a bit.
Ultimately, it would be nice to have a common base class to both Actor and Missile, and a lot of the Actor logic could be moved into this base class (say, GameObject maybe). Then the constraint of no two Actors colliding becomes no two GameObjects colliding. Missile could be a subcless of GameObject, and there could also be a Portal subclass of GameObject (it doesn't really make sense for a portal to be a Missile).
What do you think? Ofc, this is not necessary for this PR, it is alrady big enough as is.

apps/freeablo/faworld/missile/missileactorengagement.cpp
```diff
@@ -22,4 +22,26 @@ namespace FAWorld::Missile
         if (&actor != missile.mCreator)
             graphic.stop();
     }
+
+    void Missile::ActorEngagement::townPortal(Missile& missile, MissileGraphic& graphic, Actor& actor)
+    {
+        if (&actor == missile.mCreator)
```
wheybags: IIRC other players should be able to use your portal, but they can't close it.

apps/freeablo/faworld/missile/missilecreation.cpp
```diff
@@ -42,4 +44,28 @@ namespace FAWorld::Missile
...
+
+    void Missile::Creation::townPortal(Missile& missile, Misc::Point, GameLevel* level)
+    {
+        // Add portal near player
+        auto noMissilesAtPoint = [&level](const Misc::Point& p) {
+          return std::none_of(
+              level->mMissileGraphics.begin(), level->mMissileGraphics.end(),
+              [&p](const MissileGraphic* g) { return p == g->mCurPos.current();});
+        };
+        auto point = level->getFreeSpotNear(missile.mSrcPoint, INT32_MAX, noMissilesAtPoint);
```
wheybags: Not a huge deal, but I think it's better to use std::numeric_limits<int32_t>::max() than INT32_MAX

apps/freeablo/faworld/gamelevel.cpp
grantramsay: That sounds perfect.

Even with the town portal I felt it wasn't quite right. For example if you click a point directly on the other side of a portal you should get routed around it, currently you'll walk straight into it and get teleported. It should be using the "target" stuff already built in for items/actors/doors, but it's currently not that easy to do. Having a common base class (i.e. `GameObject`) with things like a "targetable" trait would make things much easier, and even help tidy up some of the existing codebase.

I agree that a portal definitely isn't a missile. It's easy to get dragged down the same implementation path as the original game when trying to re-use their assets and tables from their exe.

Yep, that'd definitely be it's own PR. I can have a look into it after this, unless you'd like to take it on yourself?

For this PR I can either remove town portal and add it back later, or just make a note that the `additionalConstraints` parameter needs to be removed after the mentioned rework is complete.

wheybags: > Yep, that'd definitely be its own PR. I can have a look into it after this, unless you'd like to take it on yourself?

If you're interested, I'm very happy to let you handle it. I When there's a provisional PR open I might add some early feedback or make some small modifications myself though.

> For this PR I can either remove town portal and add it back later, or just make a note that the additionalConstraints parameter needs to be removed after the mentioned rework is complete?

A comment is fine for now I think. Before version 1.0, having a sub-optimal feature is always better than not having it at all :)

apps/freeablo/faworld/gamelevel.h
```diff
@@ -112,9 +115,11 @@ namespace FAWorld
...
         World* getWorld() { return &mWorld; }
 
-        // These are not saved, they are added/remove in MissileGraphic constructor/destructor
-        // This avoids having to check every actor in world to find missile graphics on a level
-        std::vector<Missile::MissileGraphic*> mMissileGraphics;
+        // This list avoids having to check every actor in world to find missile graphics on a level.
+        // It is not saved, items are added/removed in MissileGraphic constructor/destructor.
+        // This is currently only intended for rendering so order is unimportant, hence using std::set
+        // and not saving/loading. Since order is not maintained this should not use be used for game logic!
+        std::set<Missile::MissileGraphic*> mMissileGraphics;
```
wheybags: We could use an std::unordered_set here instead of a normal set

wheybags:
Thanks!

Issue #478, opened 2020-03-07
Base class for all game objects


grantramsay:
Add a common abstract base class for all game objects: actors/items/doors/missiles/portals/stairs/chests/barrels etc, possibly even abstract things like traps/questTiles.

This base class should implement interfaces like:

- passable
  - can walk on/over it
  - items/openDoors/missiles/deadActors/brokenBarrels
- targetable
  - can click on it, walk to it, interact with it
  - aliveActors/items/doors/portals/stairs/unopenedChests/barrels
- hoverable?
  - highlighted and shows info text when hovered (same as targetable?)
  - aliveActors/items/doors/portals/stairs/unopenedChests/barrels

Could even be extended to include things like:

- moveable
  - whether it moves, maybe also handle movement
  - aliveActors/missiles
- interactable
  - things that aren't targetable, but something happens when an actor touches it
  - missiles, some traps, areas that alter quests
- visible
  - whether rendering needs to draw it
  - all but some missile graphics and abstract objects

This will help simplify interactions between game objects and maybe even tidy up the existing codebase a bit (e.g. consolidating existing actor/item/door/missile/stairs position maps).

Please comment with thoughts/ideas and anything I've missed or messed up.

Interesting wiki page with simple game objects discussing composition vs inheritance: https://en.wikipedia.org/wiki/Composition_over_inheritance

Original discussion: https://github.com/wheybags/freeablo/pull/475

> Ultimately, it would be nice to have a common base class to both Actor and Missile, and a lot of the Actor logic could be moved into this base class (say, GameObject maybe). Then the constraint of no two Actors colliding becomes no two GameObjects colliding. Missile could be a subcless of GameObject, and there could also be a Portal subclass of GameObject (it doesn't really make sense for a portal to be a Missile).

> Even with the town portal I felt it wasn't quite right. For example if you click a point directly on the other side of a portal you should get routed around it, currently you'll walk straight into it and get teleported. It should be using the "target" stuff already built in for items/actors/doors, but it's currently not that easy to do. Having a common base class (i.e. GameObject) with things like a "targetable" trait would make things much easier, and even help tidy up some of the existing codebase.

wheybags:
> This base class should implement interfaces like:

[...]

PR #479, opened 2020-03-08
Fix/save const correctness

grantramsay:
This is a single commit appended on same branch as #475 (because of conflicts), the diff will make more sense after that is merged.
Just added `const` to all `save` methods I could find.
Adding the `save` method to a common base class (#478) requires that items/actors etc have the same `save` method signature, and it'd be a bit messy to include this in that PR.
Things that requiring saving should probably be implementing an `ISaveable` interface, but that can be done another day


wheybags:
Looks good, thanks!

PR #481, opened 2020-03-13
graphics refactoring

wheybags:
Adding a C++ wrapper for opengl, whose API is (somewhat) inspired by vulkan. This should ease porting later on. Doesn't much address the mess that is sdl2backend.cpp, that will be the next PR.


grantramsay:
Awesome, looking good!

The GUIs didn't show up on my mac, adding `layout(location =` in the nuklear vertex shader (same as you did for other shader) seems to fix

wheybags:
@grantramsay just to be sure, could you maybe try running the game on osx again, to be sure I didn't break anything?

grantramsay:
@wheybags - everything but the highlight/hover outside of GUI seems to be working great.

For a quick test I changed `SpriteVertexPerInstance::v_hoverColor` to `int16_t` and format to `Format::RGBA16I`, and highlight/hover started working again, but I'm not sure of root of root cause

grantramsay:
Also the atlas texture fills up after ~4 levels and hits an assert.
Looks like a change to `PlayerBehaviour::addInput()` `case PlayerInput::Type::ChangeLevel` was dropped in a merge. `mPlayer->getWorld()->setLevel()` clears the atlas texture when the player changes level, but this was reverted back to `mPlayer->teleport()`

wheybags:
Thanks, fixed both of these issues here: https://github.com/wheybags/freeablo/pull/482
For the atlas, my plan is to compute bounding boxes of the _actually used_ pixels, and only put those parts in the atlas. It will increase the renderer complexity quite a lot, but I think it's worth it when you look at the actual atlas usage:
image

That was the atlas after just loading the town. Green is unused. As you can see, almost all of it is wasted space. The level tiles are huuuuge, but only use a tiny square at the bottom. On top of that, a lot of them are just literally empty.

grantramsay:
Wow yeah that's a lot of wasted space in those textures.
Sound's like a good plan, especially since reloading sprites between level changes is so slow.
Maybe have a look at the atlas image a few levels down in the dungeon too, IIRC I was surprised how much room the monster textures used. Maybe doing the palettes in shader would help too?

PR #483, opened 2020-03-26
Trim atlas sprites

wheybags:


wheybags:
Trimming the atlas works quite well:
image
(This is with all levels loaded)

grantramsay:
Awesome! Looks a bit like the matrix 😂
Is that the same bin packing algorithm, the packing looks a bit strange?
Is that 2 layers of 8k*8k textures in that image?

wheybags:
Yup, it's the same one. Packing is sub optimal I think because of the random order images get inserted in. I think if I sort the frames by size before inserting, it should dramatically increase utilisation.
And yep, it's the two 8k layers on top of eachother, I used https://apitrace.github.io/ to capture (it's a great tool, if you haven't tried it before)

PR #486, opened 2020-04-24
Graphics refactor 4

wheybags:


AJenbo:
in case you are having the issue (i belive i saw it in a video about freeablo) where players and monsters poke through walls as they walk alongst walls in the lower left of them. the way to solve that is to first render non solid tiles for a given row, then render the solid tiles for that row.

wheybags:
@AJenbo I'm not sure what you mean by "solid tiles"?

AJenbo:
Solid tiles are tiles that the players and monsters can not walk on, so walls, lave, rocks etc.

The reason is that when rendering left to right, tile by tile, an offset sprite (moving player) can overlap part of a wall that had previously been rendered.

Afaik this is only a problem if you render tile by tile like the original render, if you instead layout all sprites (to get proper sorting) and then render them, this shouldn't be an issue. I'm not aware of how your current render works, so if you have made this change you might already have solved the issue.

By the way did you also figure out how to render trees in town correct?
tree

In the original render, there is also another issue where some parts of sprites get obscured by the next row of ground tiles, the fix is similar to the aforementioned one, except you have to render all ground tiles (and only the diamond shape) before rendering anything else.
render1

Let me know if Freeablo is affected by any of these issues and I can help detail it further.

wheybags:
@AJenbo ahh, ok.
We do have bugs with ordering somewhat. We currently draw all ground diamonds, then draw the walls + entities north-northeast axis major fashion. I would like to fix the issues, but it's good enough for now. In freeablo's case, for things like the trees, if we could combine it into one tree sprite, then we can just sort items by the y value of the bottom of the sprite also.
I also have a pipe dream of maybe one day making some simple wall meshes and actually drawing the levels as 3d scenes, so in that case we could use the proper z buffer to order them.

PR #493, opened 2020-05-21
regarding issue #491- pepin healing

neveza:
for issue #491 - added heal function and sound for pepin heal option. Also added a heal function with int32_t parameter to specify how much health, mostly for debugging, but may be useful for potions and spells future implementation

### DIFF
--- a/apps/freeablo/faworld/actor.cpp
+++ b/apps/freeablo/faworld/actor.cpp
@@ -223,6 +223,7 @@ namespace FAWorld
     }
 
     void Actor::heal() { mStats.getHp().current = mStats.getHp().max; }
+    void Actor::heal(int32_t toHeal) { mStats.getHp().current += toHeal;}
 
     void Actor::restoreMana() { mStats.getMana().current = mStats.getMana().max; }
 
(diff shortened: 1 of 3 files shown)


apps/freeablo/faworld/actor.cpp
```diff
@@ -223,6 +223,7 @@ namespace FAWorld
     }
 
     void Actor::heal() { mStats.getHp().current = mStats.getHp().max; }
+    void Actor::heal(int32_t toHeal) { mStats.getHp().current += toHeal;}
```
vieiraa, line 226:
```suggestion
    void Actor::heal(int32_t toHeal) { mStats.getHp().current += toHeal; }
```

wheybags:
Hi @neveza, thanks for the PR! I made one small tweak to the Actor::heal() function. I changed it to call MaxCurrentItem::add, as this will make sure it stays in bounds. Previously, the function could heal you to over max health.

Also, you may have noticed that the continuous integration failed: https://travis-ci.org/github/wheybags/freeablo/builds/689485091?utm_source=github_status&utm_medium=notification

It was just a minor formatting faliure, if you click into the clang-format job that failed and look at its output, you can see that it provides a diff to automatically fix the problem (but it's fine this time, my change ovewrote the bad line anyway).

Issue #492, opened 2020-05-16
Use potions


wheybags:

neveza:
Currently working on this one. I've implemented healing, mana, rejuvination, and the full equivalents of the potions except for elixers.

wheybags:
Nice! Feel free to open a PR even if it's not finished yet, I won't merge until you let me know it's done.

wheybags:
Also, just so you know, I'm currently working on a refactor of the item handling code (at #495), so don't worry too much about hacking around the not-so-nice item class.

neveza:
While working on the elixirs, I may have gone a bit above and beyond. I noticed the baseStats weren't using MaxCurrentItem such like Hp/Mana, so I decided that instead of repeatedly implementing switch/ifs checks, I updated the BaseStats to use MaxCurrentItem as I figured it'd help in the long run while it's early.

So, 
        instead of
 `       int32_t strength = 0;`

       it is this
        `mutable Misc::MaxCurrentItem<int32_t> strength;`

then code such as:
  ` stats.baseStats.strength.current = charStats.strength.current +itemStats.baseStats.strength.current;`
now won't go out of character maxes
  ` stats.baseStats.strength.add(charStats.strength.current + itemStats.baseStats.strength.current);`

I've done a rough dirty patch up to see if I didn't break anything obvious. So far, nothing seems out of the norm. That said, if this is disagreeable, I will revert back, otherwise I can continue seeing if anything can be updated appropriately then commit and pull request.

wheybags:
I'm not quite sure what you're trying to do with that? IIRC attributes don't have a current/max status like life and mana? Also, why mutable?

neveza:
Attributes has a max. You have the base max and then the max after any bonuses from equipment if factored in. Each character has their own specific maxes. http://www.bigd-online.com/JG/JGFrame.html ; so doing this, would allow the max restriction for the base, then for the livestats, you'd have the 'after' equipment bonuses are factored in with those maxes. Hence why I went this way because otherwise, you'll have to not only account for base restriction, but the after equipment bonus which means a lot of if conditions.

As for the mutable, I'm just following suite with what was there. After all, this has to be mutable data, but unsure if it was needed or not.

wheybags:
[...]

PR #496, opened 2020-05-25
Potions and Elixers #492

neveza:
for #492, All potions and elixirs added along with functions to add to attributes


apps/freeablo/faworld/potion.h
```diff
@@ -0,0 +1,26 @@
...
+#include "player.h"
+
+namespace FAWorld
+{
+    // Class Potion for all potion functions
+    class Potion
+    {
+    public:
+        Potion();
+        ~Potion();
```
wheybags, line 12: This class doesn't need to have a constuctor or destructor, the methods could all be made static, or it could turned into a namespace instead.

apps/freeablo/faworld/potion.cpp
```diff
@@ -5,33 +5,31 @@
...
         switch (player->getClass())
         {
             case FAWorld::PlayerClass::warrior: {
-                bonus = 2.0;
+                bonus = "2.0";
                 break;
             }
             case FAWorld::PlayerClass::rogue: {
-                bonus = 1.5;
+                bonus = "1.5f";
```
wheybags, line 21: These should all be FixedPoint expressions, not strings, eg: `FixedPoint bonus = 1; ... bonus = FixedPoint("1.5");`

neveza:
I'm unsure of the failures regarding appveyor/travis-ci. Seems it isn't seeing the potion.h/.cpp files to link the functions after I changed all functions to static. Visual studios compiles just fine however. My inexperience with this process has left me in the dark, unfortunately.

apps/freeablo/fagui/guimanager.cpp
```diff
@@ -189,6 +191,116 @@ namespace FAGui
                 mGoldSplitCnt = 0;
                 break;
             }
+            case FAWorld::ItemType::misc:
+            {
+                switch (item.getMiscId())
+                {
+                    case FAWorld::ItemMiscId::potionOfHealing:
+                    {
+                        FAWorld::Potion::restoreHp(mPlayer);
```
wheybags, line 200: In order to work in multiplayer, this code needs to be changed so it sends the action to the server, and then only actually performs the healing when the server sends the action back out to all clients.
The player input types are here: https://github.com/wheybags/freeablo/blob/master/apps/freeablo/faworld/playerinput.h you'll need to add a new one. They get processed here: https://github.com/wheybags/freeablo/blob/master/apps/freeablo/faworld/playerbehaviour.cpp#L70. Here's an example of sending one: https://github.com/wheybags/freeablo/blob/master/apps/freeablo/engine/localinputhandler.cpp#L29. You can send an EquipTarget as part of the event, that stores the location in inventory, and then on the ither end you just get the item from that location, make sure it's stipl there, and use it. Sorry I didn't notice before now.

wheybags:
Hey, just wanted to let you know that I merged a bunch of item refactoring code, so there's conflicts now because of that. Don't worry about them though, I can handle the merging once this branch is ready.

neveza:
Hopefully, this works for the multiplayer/network aspect. I've also updated the cmakefile to add the .h/.cpp for potion.

wheybags:
Thanks!

Issue #399, opened 2018-11-10
Wirt's dialog runs off screen


wheybags:

neveza:
Currently working on this. So far, I've managed to make it very close to how it looks in Diablo. That said, not too familiar with nuklear, so going to see if I cannot make make things a bit more cohesive code wise.
![image](https://user-images.githubusercontent.com/5185373/85224221-b0267c80-b396-11ea-8f30-95dee122433d.png)

PR #498, opened 2020-06-24
Text scroll bug fix

neveza:
Fix for the #399 bug. All dialog text boxes don't run off the background frame now.


vieiraa:
I think this PR is including some changes from your previous PR. Would you be able to rebase?

PR #500, opened 2020-09-12
Add missing headers

vieiraa:
Clang was failing to compile.


AJenbo:
It looks like you changed the linen endings, you probably want to avoid that so that the diff gets easier to read

vieiraa:
Yeah, Visual Studio/Windows sucks. Will be careful next time.

